PR #119, opened 2022-03-25
Add information on how to change the start address

sebromero:
## What Needs To Be Reviewed
- If the added information is correct

## How To Give Feedback
Please leave your feedback as a [Github review](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews). 
You can add comments to specific lines of content / code and ideally use Github's [suggestion feature](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request). 🙏


sebromero:
@facchinm LGTY?

facchinm:
Apart from modifying the json, the user should also recompile using `mbed-os-to-arduino` script; otherwise the "normal" method would be to change the .text section start here https://github.com/arduino/ArduinoCore-mbed/blob/master/variants/PORTENTA_H7_M7/linker_script.ld#L3

PR #121, opened 2022-03-25
Fix XHTML issue in datasheet

sebromero:


sebromero:
@marqdevx FYI

marqdevx:
Oops always forget about that!!
Maybe we could add it on the linter :thinking: 
thanks :ok_hand:

PR #293, opened 2022-06-14
Fix difficulties [PC-932]

marqdevx:
## What This PR Changes
- Remove wrong difficulty labels

PR #294, opened 2022-06-14
Add fixed strings for the difficulty tag [PC-938]

marqdevx:
## What This PR Changes
- The Linter tool now will only accept the following difficulty tags: `Easy`, `Intermediate`, `Advanced`

### DIFF
--- a/scripts/validation/rules/tutorial-metadata-schema.json
+++ b/scripts/validation/rules/tutorial-metadata-schema.json
@@ -30,7 +30,8 @@
         "type": "string"
       },
       "difficulty": {
-        "type": "string"
+        "type": "string",
+        "enum": ["Easy", "Intermediate", "Advanced"]
       },
       "hero_position": {
         "type": "integer"
--- a/scripts/validation/rules/tutorial-metadata-schema.json
+++ b/scripts/validation/rules/tutorial-metadata-schema.json
@@ -31,7 +31,7 @@
       },
       "difficulty": {
         "type": "string",
-        "enum": ["Easy", "Intermediate", "Advanced"]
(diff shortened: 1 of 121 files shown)


marqdevx:
It would be nice to get first what is happening here #293

scripts/validation/rules/tutorial-metadata-schema.json
```diff
@@ -30,7 +30,8 @@
         "type": "string"
       },
       "difficulty": {
-        "type": "string"
+        "type": "string",
+        "enum": ["Easy", "Intermediate", "Advanced"]
```
sebromero, line 34:
```suggestion
        "enum": ["beginner", "intermediate", "advanced"]
```
I think the more common range is beginner - intermediate advanced. Shall we go for that instead? Or otherwise easy medium hard.

sebromero (dismissed):
Thanks Pablo! Please have a look at my comment.

scripts/validation/rules/tutorial-metadata-schema.json
marqdevx: Sounds good to me, @karlsoderby what do you think? 👀

sebromero: Once we make a decision, we need to change them actually in the same PR, otherwise the linter will complain :-)

karlsoderby: Hey @marqdevx so the idea is to change `easy` > `beginner`? Sounds good to me, it is a better label.

marqdevx:
I will make the needed changes on the other PR, and then merge it here
#293 into #294

sebromero (approved):
LGTM

PR #295, opened 2022-06-15
Make clear that you cannot transmit and receive data at the same time

yaxu:
More info: https://stackoverflow.com/questions/39436771/garbage-data-received-if-transmiting-some-data-from-the-same-softwareserial-port

The lack of this info lost me a couple of days. I'm surprised that it's not made clear in the documentation or in the examples.


marqdevx (approved):
LGTM

PR #410, opened 2022-07-25
Update assets install library IDE 2 article

ArduinoBot:
## What This PR Changes
- Update assets install library IDE 2 article

## Contribution Guidelines


per1234 (changes requested):
## General

The readability of the screenshots has become significantly worse due to the IDE window having been sized unnecessarily large when taking them. This makes the text and interface difficult to see in the final image.

Please make your IDE window smaller and then redo the screenshots so the scale in the final image is larger. The scale in the previous screenshots was acceptable.

## `content/software/ide-v2/tutorials/ide-v2-installing-a-library/assets/installing-a-library-img03.png`

I think the previous approach of showing the in progress state of the installation was better.

The IDE does this via a notification like this:

image

## `content/software/ide-v2/tutorials/ide-v2-installing-a-library/assets/installing-a-library-img06.png`

I recommend uninstalling some libraries so the **File > Examples** menu is not quite so huge.

I also think the dark OS theme harms readability here, but that is fairly subjective and I am biased because I'm not a fan of dark themes in general.

per1234 (approved):

PR #460, opened 2022-08-13
Capitalization and ITS

MessiahJa:
Hi,
I just capitalized the first letters of the last set of bullet points and changes *it's* to *its*
## Contribution Guidelines


marqdevx:
They are not caps because they are separated by commas, so it is like a normal numeration.

marqdevx (approved):
I've fixed the commit to only have the 'its' typo

Thanks for the PR!

PR #501, opened 2022-08-23
Scripts: Spell-check for windows

marqdevx:
## What This PR Changes
- Add spell-check for windows

### DIFF
--- a/scripts/spell-check.cmd
+++ b/scripts/spell-check.cmd
@@ -1,5 +1,13 @@
 @ECHO OFF
 
+pip --version || pip3 --version
+IF %ERRORLEVEL% NEQ 0 (
+    cls
+    ECHO Please install Python from https://www.python.org/downloads/ and Pip following https://pip.pypa.io/en/stable/installation/
+    EXIT /B 
+)
+
+cls
 codespell --version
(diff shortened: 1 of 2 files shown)


scripts/spell-check.cmd
```diff
@@ -0,0 +1,20 @@
...
+codespell --version
+IF %ERRORLEVEL% NEQ 0 (
+    ECHO Installing codespell
+    pip install codespell || pip3 install codespell
+) else (
+    ECHO Codespell already installed
+)
+cls
+codespell -I scripts/resources/spell-check-ignore-list.txt --skip="*.svg,*.dxf" content
+ECHO Spell-check finished
```
sebromero, line 20:
```suggestion
```
I'd keep the output the same as with the Unix version. VSCode should indicate when the task is done, so imo no extra message needed.

sebromero (changes requested):
Almost ready to go :-)

scripts/spell-check.cmd
marqdevx: Oops, didnt look at the unix version echo :S
Removing it

sebromero (approved):
LGTM

Issue #134, opened 2022-04-03
Possible wrong RGB sequence


tahunus:
In Line 128 of Arduino_MKRIoTCarrier.h the LED instance is created with: 
` Adafruit_DotStar leds = Adafruit_DotStar(NUMPIXELS, DATAPIN, CLOCKPIN, DOTSTAR_BGR); `
(link to file : [https://github.com/arduino-libraries/Arduino_MKRIoTCarrier/blob/master/src/Arduino_MKRIoTCarrier.h](url))

Which suggests the color sequence is BGR, not RGB. The Adafruit Neopixel library in use, does not have a BGR sequence, but it does have a GRB sequence: [https://learn.adafruit.com/adafruit-neopixel-uberguide/arduino-library-use](url)

In any case , the MKR IoT Carrier Cheat Sheet has an error in 
[https://docs.arduino.cc/tutorials/mkr-iot-carrier/mkr-iot-carrier-01-technical-reference#leds](url)  where the picture lists how to get the colors using the RGB order [https://docs.arduino.cc/static/e66517362e06a9efb932279bbf51e6dc/29114/mkrIotCarrier-leds.png](url)

Then the code below it suggest the sequence green, red, blue:
`carrier.leds.setPixelColor(index, green, red, blue);`

Which works correctly, although it's not the BGR of line 128 of MKRIoTCarrier.cpp

Total confusion

marqdevx:
We will make a PR on the MKRIoTCarrier library to make the LEDs be RGB as it is more common.

marqdevx:
Update: Library is OK
I will make changes on our documentation to show the correct order, which is RGB

marqdevx:
Thanks for spotting the issue @tahunus 🙌 
It should be fixed now 😄

PR #516, opened 2022-08-24
MKRIoTCarrier tech reference: Fix RGB from #134

marqdevx:
## What This PR Changes
- MKRIoTCarrier tech reference: fix LED's RGB command reference from #134


karlsoderby (approved):
LGTM @marqdevx

PR #534, opened 2022-09-03
fixed typos -- see commit description

omajoshi:
whereas is incorrect in this context (and supports->support should agree with the plural "pins") -- I think "12 of which" is more appropriate but another option is "The Due has 54 digital pins, including 12 which support PWM (Pulse Width Modulation)."

## What This PR Changes
- typos, including subject-verb agreement

## Contribution Guidelines


karlsoderby (approved):
Thank you @omajoshi for this correction.

PR #535, opened 2022-09-05
Benjamindannegard/ide2.0 debuging screenshot updates

BenjaminDannegard:
## What This PR Changes
- Updated screenshots for MKR debuging tutorials

## Contribution Guidelines


per1234 (changes requested):
The new screenshots use an version of the Arduino IDE that is outdated by over 8 months.

Some significant changes have been made to the relevant parts of the Arduino IDE Arduino IDE UI since the version of the IDE used to create these screenshots.

Please update to Arduino IDE 2.0.0-rc9.3 and do them over again.

per1234 (approved):
Looks good. Thanks @BenjaminDannegard 

Ideally, those four screencasts at the end of the Atmel ICE tutorial would also be updated, but I think the most important part was the initial screenshots that introduce the user to the interface.

After that, they will at least be less likely to be confused that the interface in the screencast looks different from the IDE they are using.

PR #575, opened 2022-10-06
UNO Rev3 Pinout revamp

FunzioneFabio:


sebromero (changes requested):
This file was uploaded to the root folder, it won't work

jhansson-ard:
@sebromero @FunzioneFabio was this change implemented somewhere else and can we delete this branch? 🙏🏼

FunzioneFabio:
@jhansson-ard I see that it has been updated both in the store and in docs, we can close this.

PR #581, opened 2022-10-07
MKR Relay Shield: Fix dimensions [TC-365]

marqdevx:
## What This PR Changes
- Fix dimensions of  the MKR Relay Shield


jhansson-ard (approved):
LGTM

PR #617, opened 2022-10-27
Niclas tech-specs: Add PCB Thickness and align units format

marqdevx:
Update Tech-specs:
* Add PCB Thickness
* Align units format


canchebagur (approved):
LGTM!

PR #678, opened 2022-11-22
[PC-1042] Opta family block on docs page

jcarolinares:
## What This PR Changes
- Opta family block

## Contribution Guidelines


marqdevx:
Could the name of the folder be `pro-plc` then inside can be the opta folder?

jcarolinares:
> 

It's not a bad idea but I'm afraid pro-plc will be confusing in the long term, the idea was to have a block only for the Opta family.

PR #681, opened 2022-11-22
Action: Datasheets render

marqdevx:
## What This PR Changes
- New GitHub Action to render the datasheets, and check if there are issues on the rendering

I did sometesting at https://github.com/marqdevx/docs-content-private/actions/runs/3524139452
(private repo)

### DIFF
--- a/scripts/datasheet-rendering/package.json
+++ b/scripts/datasheet-rendering/package.json
@@ -1,6 +1,6 @@
 {
   "name": "datasheet-rendering",
-  "version": "1.0.0",
+  "version": "1.0.2",
   "description": "",
   "type": "module",
   "main": "render-datasheets.js",
--- a/scripts/datasheet-rendering/package.json
+++ b/scripts/datasheet-rendering/package.json
@@ -1,6 +1,6 @@
 {
   "name": "datasheet-rendering",
-  "version": "1.0.2",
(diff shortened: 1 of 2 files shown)


scripts/datasheet-rendering/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "datasheet-rendering",
-  "version": "1.0.0",
```
sebromero, line 3: This doesn't refer to the datasheet-renderer package and hence has no effect here.

sebromero:
@marqdevx Looks good, what's the purpose though?

scripts/datasheet-rendering/package.json
marqdevx: True
What is that referring to?

sebromero: The script that integrates the renderer is also a node module. One that's not published in the registry but still it's a module.

scripts/datasheet-rendering/package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "1.0.2",
```
sebromero, line 3:
```suggestion
  "version": "1.0.0",
```

sebromero (approved):
LGTM

PR #717, opened 2022-12-07
Update analog-output.md

topquark22:
## What This PR Changes

## Contribution Guidelines

### DIFF
--- a/content/learn/02.microcontrollers/03.analog-output/analog-output.md
+++ b/content/learn/02.microcontrollers/03.analog-output/analog-output.md
@@ -13,6 +13,8 @@ In the graphic below, the green lines represent a regular time period. This dura
 
 ![](./pwm.gif)
 
+PWM is only available on digital output pins 3, 5, 6, 9, 10 and 11. The other digital pins cannot be used for PWM.
+
 Once you get this example running, grab your Arduino and shake it back and forth. What you are doing here is essentially mapping time across the space. To our eyes, the movement blurs each LED blink into a line. As the LED fades in and out, those little lines will grow and shrink in length. Now you are seeing the pulse width.
 
 *Written by Timothy Hirzel*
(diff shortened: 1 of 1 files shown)


karlsoderby:
Hello @topquark22 , thank you for your submission. However, this only applies to some boards (particularly AVR boards). For example, the Nano RP2040 supports PWM on 20 pins (basically all pins).

content/learn/02.microcontrollers/03.analog-output/analog-output.md
```diff
@@ -13,6 +13,8 @@ In the graphic below, the green lines represent a regular time period. This dura
 
 ![](./pwm.gif)
 
+PWM is only available on digital output pins 3, 5, 6, 9, 10 and 11. The other digital pins cannot be used for PWM.
```
sebromero, line 16:
```suggestion
On some microcontrollers PWM is only available on selected pins. Please consider the pinout diagram of your board to find out which ones you can use for PWM. They are denoted with a tilde sign (~).
```

sebromero (changes requested):
I made a proposal. Please let me know what you think.

sebromero:
@karlsoderby What do you think about my suggestion?

content/learn/02.microcontrollers/03.analog-output/analog-output.md
karlsoderby: @sebromero yes this works great, thanks. Let's see if in a later stage, we can add visuals to explain this better.

sebromero:
@topquark22 I'm referring to the pinout diagrams that we provide. There we use the tilde symbols also for nano.

ABX00031-pinout

A000005-pinout

topquark22:
Another thing you should mention is the frequency of the PWM signal. I measured it on my scope (from a Nano) and it was 976Hz.
G
----

    On Wednesday, December 7, 2022 at 10:24:43 a.m. EST, Sebastian Romero ***@***.***> wrote:  
 
 


@topquark22 I'm referring to the pinout diagrams that we provide. There we use the tilde symbols also for nano.

sebromero:
I guess the PWM frequency again depends on the board, so I think I wouldn't add it in this generic article.

PR #730, opened 2022-12-14
IoT API Article changes

eclipse1985:
## What This PR Changes
- We have only 2 API types, so the article has been restructured a bit
- The original mean was to be a sort of index, so API details should go in other places

## Contribution Guidelines

### DIFF
--- a/content/arduino-cloud/01.getting-started/04.api-overview/api-overview.md
+++ b/content/arduino-cloud/01.getting-started/04.api-overview/api-overview.md
@@ -2,7 +2,7 @@
 title: Arduino IoT Cloud APIs
 description: A summary of the three APIs available in the Arduino IoT Cloud
 author: Karl Söderby
-tags: IoT Cloud, API, JavaScript
+tags: [IoT Cloud, API, JavaScript]
 ---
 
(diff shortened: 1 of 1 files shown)


alranel (changes requested):
“Applications” is a bit confusing to me because it seems to apply to both APIs, no? Personally, I would use “Configuration API” and “Data API” to base the name on the actual purpose and distinguish them better…

I would also improve a bit the structure of the article which is not super clear. For instance, the “backward compatibility policy” only applies to the REST API and not to the MQTT one.

Proposed structure:
* One-sentence intro “Arduino IoT Cloud has two different set of APIs. Read below to learn more”.
* Configuration API
   * What you can achieve with it (practical examples, please avoid generic sentences like “Automate and integrate different IoT Solutions with a single, powerful Cloud”)
   * What resources/actions are available through it
   * Link to documentation
* Data API
   * What you can achieve with it (practical examples)
   * What resources/actions are available through it
   * Link to documentation

I believe backward compatibility policy should be moved to the documentation page of each API or library, because they are different.

(Sorry for being so pedantic but I care a lot about documentation clarity)

alranel (approved):

PR #759, opened 2023-01-11
Fixed erroneous example.

jfjlaros:
## What This PR Changes
This PR addresses two issues with an example in the tutorial style guide:
- `TRUE` is not defined.
- Redundant comparisons in conditions are [not recommended](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es87-dont-add-redundant--or--to-conditions).

## Contribution Guidelines

### DIFF
--- a/content/learn/08.contributions/00.arduino-writing-style-guide/arduino-writing-style-guide.md
+++ b/content/learn/08.contributions/00.arduino-writing-style-guide/arduino-writing-style-guide.md
@@ -53,14 +53,14 @@ Put your `setup()` and your `loop()` at the beginning of the program. They help 
 Use verbose if statements. For simplicity to the beginning reader, use the block format for everything, i.e. avoid this:
 
 ```
-if (somethingIsTrue) doSomething;
+if (distance > 10) moveCloser();
 ```
 
 Instead, use this:
 
 ```
-if (somethingIsTrue) {
-   doSomething;
+if (distance > 10) {
+   moveCloser();
 }
 ```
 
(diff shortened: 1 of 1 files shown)


content/learn/08.contributions/00.arduino-writing-style-guide/arduino-writing-style-guide.md
```diff
@@ -59,7 +59,7 @@ if (somethingIsTrue) doSomething;
 Instead, use this:
 
 ```
-if (somethingIsTrue == TRUE) {
+if (somethingIsTrue) {
```
karlsoderby, line 62:
```suggestion
if (somethingIsTrue == true) {
```

karlsoderby: Hey @jfjlaros , I suggest that we keep the `== true` in this if statement, as for less experienced coders, it describes what is happening a bit better.

While `if(somethingIsTrue){}` works the same and involves less code, it is harder to understand for a beginner, as the state it is compared (true) is not clear. If you want to, we can add this as a paragraph below.

Thanks a lot for your submission, let me know what you think.

jfjlaros: I understand your concerns, but I am of the opinion that we should not encourage this practice. As mentioned in the C++ core guidelines, it is redundant and opens opportunities for errors. The current issue is a good example of this.

I also think that adding redundant truth tests will get the wrong message across for beginners, as they may get the impression that `if` is some sort of function that takes two operands.

As a compromise, maybe the example could be changed to one that uses a meaningful comparison, e.g.,

```cpp
if (distance > 10) {
  // Go nearer.
}
```

Then we could leave the choice of using `== true` up to the author of the tutorial.

karlsoderby (approved):
Thank you @jfjlaros, I think this looks good.

PR #807, opened 2023-01-26
Correction of udev rules in the Ubuntu Chapter

frigi83:
I have troubled with this problem for 3 month. But after a step by step debug I have found a solution. 

The two steps with the text for the udev rule were both incorrect and the rule was not applied.

You can debug the udev tree with this command: `sudo udevadm info -a -p $(udevadm info -q path -n /dev/ttyACM0)` (asuming that your device is mounted as `/dev/ttyACM0`).

I am not shure if it is the only page were this suggestion is present.

## What This PR Changes
- it correct the udev rules in the chapter ***Ubuntu Issues Troubleshooting***, both steps contains errors

## Contribution Guidelines


mcmchris (commented):
This looks good to me, but I don't have Ubuntu to test this. 

cc. @jcarolinares @TaddyHC

PR #866, opened 2023-02-28
updated link for OPTA

akash73:
## What This PR Changes
added https://docs.arduino.cc/hardware/opta

## Contribution Guidelines


TaddyHC (dismissed):
Thanks for the extension, lgtm! @jcarolinares , I think it would be nice if you could take a quick look at it too. It's regarding the added link to Opta documentation page instead of the store link (Which we haven't included since it could be shared with Finder) for the updated Arduino & Modbus article. I think it should be good but checking in just in case.

jcarolinares (dismissed):
It's a nice solution for the store link problem, but, we have to wait, the link won't work and if possible I would move Opta to a new hardware block

akash73:
@TaddyHC @jcarolinares not sure I've understood why is blocked. in the page there's no mention to any store link.
every product listed before the OPTA has a link do docs, not to the store.

akash73:
up

jcarolinares (approved):
Unblocked

Looks good to me

Merging

PR #931, opened 2023-03-21
Bauds per second to baud rate

alrvid:
## What This PR Changes
Baud is a unit for symbol rate, which means that time is already taken into account (symbols per second). So “bauds per second” sounds like a unit of symbol rate acceleration, which is incorrect.

## Contribution Guidelines
I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


sebromero (approved):
LGTM

PR #936, opened 2023-03-21
Made information about latency clearer

alrvid:
The explanation is rewritten with the user's perspective in mind - avoiding, for example, terms like FIFO and interrupt. Also, it now makes explicit what a latency of 0 means.

I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


sebromero (approved):
LGTM

PR #938, opened 2023-03-21
Fixed quaternion syntax example error

alrvid:
There’s a syntax example that shows how to print the quaternion rotation scalar, but it mistakenly uses the orientation object instead of the quaternion object.

[x] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


sebromero (approved):
LGTM

PR #955, opened 2023-03-29
match PIN numbers to colors

zhumo:
## What This PR Changes
Hello, I was going through the tutorial on RGB for the makr 1010 unit and found that the marked numbers in the docs don't match the actual pins. Corrected in this PR. 

## Contribution Guidelines

### DIFF
--- a/content/hardware/01.mkr/01.boards/mkr-wifi-1010/tutorials/built-in-rgb/built-in-rgb.md
+++ b/content/hardware/01.mkr/01.boards/mkr-wifi-1010/tutorials/built-in-rgb/built-in-rgb.md
@@ -56,7 +56,7 @@ Controlling the RGB on the MKR WiFi 1010 is slightly different from any previous
 #include <utility/wifi_drv.h>
 ```
 
-After we have included it, it is pretty straightforward. The RGB component uses 25 (green), 26 (red) and 27 (blue), so we need to define those pins by using the following commands:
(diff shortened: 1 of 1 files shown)


karlsoderby:
Hey there @zhumo ! So just to clear up some confusion, the examples provided in this tutorial are in fact correct. The wiring does not follow the classic **RGB** order, but instead uses **GRB**.

When I created this example, I listed out the order of pins as opposed to the colors. This means that:
- Pin 25 = G
- Pin 26 = R
- Pin 27 = B

So unfortunately, the changes you are suggesting are not correct. However, it is perhaps poorly explained from our side, so I will add a clarification. Thanks a lot for reporting!

zhumo:
@karlsoderby thanks for your quick reply! 

Are you sure, though? I ran this code on my MAKR WiFi 1010:

```
#include <WiFiNINA.h>
#include <utility/wifi_drv.h>

void setup() {
  WiFiDrv::pinMode(27, OUTPUT);
  WiFiDrv::digitalWrite(27, HIGH);
}

void loop() {
  // NOOP
} 
```

And I replaced the `27` with 25 and 26. When I did that, I got that 25 is red, 26 is green, and 27 is blue... Has the hardware changed or something? This is the one I bought https://www.amazon.com/Arduino-MKR-WiFi-1010-ABX00023/dp/B07FYFF5YZ/ref=sr_1_2?crid=RXTJX56C38IW&keywords=arduino+maker+wifi&qid=1679380916&sprefix=arduino+maker+w%2Caps%2C215&sr=8-2

Anyway, not a big... definitely not the end of the world if the tutorial said something different.

PR #1095, opened 2023-06-14
Update README.md

Matteo-it:
## What This PR Changes
Update README to explain CodeBlocks elements


jhansson-ard (commented):
This looks good to me! However I want to tag @karlsoderby, @jacobhylen and @jcarolinares  here as well to hear their thoughts, as they will use this most!

CloruroDiSodio (approved):

karlsoderby:
This looks great, exactly what we expected @Matteo-it

As for the `language` parameter, is it possible to see which ones are supported?

Matteo-it:
> This looks great, exactly what we expected @Matteo-it
> 
> As for the `language` parameter, is it possible to see which ones are supported?

I didn’t specify it because it supports the same languages that are already supported now, so you can put in the class what you currently place in the markdown!

PR #1166, opened 2023-07-10
Update cheat-sheet.md

V205Github:
Fixed broken. Link?

## What This PR Changes
- (Please explain here why you created the pull request and specify what it changes
- The  link for the RA4M1 datasheet was broken. 
- Could somebody confirm that the link I added to the docs works? Should it be changed to the specific part instead of family?
Thank you.

## Contribution Guidelines

### DIFF
--- a/content/hardware/02.hero/boards/uno-r4-minima/tutorials/cheat-sheet/cheat-sheet.md
+++ b/content/hardware/02.hero/boards/uno-r4-minima/tutorials/cheat-sheet/cheat-sheet.md
@@ -17,7 +17,7 @@ software:
   - ide-v2
 ---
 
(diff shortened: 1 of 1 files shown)


Hannes7eicher:
Thank you for pointing out the missing link. We host the datasheet pdf on in our repository instead of linking to the URL. The issue was that the file for some reason was missing from our repository so even though the link was correct the file could not be displayed. I've fixed the issue in this [PR](https://github.com/arduino/docs-content/pull/1167) so I will close your PR because we're using a different way of linking to the datasheet but otherwise good catch and thanks for the PR.

V205Github:
Thank you!

PR #1173, opened 2023-07-12
Update string title and array source to fix type definition in docs

Matteo-it:
## What This PR Changes
Developing docs 2.0.0 we found some inconsistency with type definition for **title** and **source** frontmatter:
- Title should be a string, in one case we have a number;
- Source it's unused in docs (i guess), but anyway we managed it as a string. We have one case where we have an array.

## Contribution Guidelines

### DIFF
--- a/content/hardware/_unlisted/boards/101/product.md
+++ b/content/hardware/_unlisted/boards/101/product.md
@@ -1,4 +1,4 @@
 ---
-title: 101
+title: "101"
 certifications: [CE]
 ---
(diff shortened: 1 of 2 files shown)


jhansson-ard:
@Matteo-it do we need some manul updates of the tutorials as well to check that all titles in  the frontmatter are strings?

Matteo-it:
> @Matteo-it do we need some manul updates of the tutorials as well to check that all titles in the frontmatter are strings?

@jhansson-ard No, I saw that the 2 files I edited are the only two that create type warnings. Anyway it works with number title and array source, but is better remove the easy warning now that we are upgrading to docs 2.0 :)

jhansson-ard (approved):
LGTM

PR #1186, opened 2023-07-14
[MKC-1009] UNO R4 WiFi Qwiic tutorial

jacobhylen:
## What This PR Changes
- Adds a tutorial for how to use the Qwiic connector on the Arduino UNO R4 WiFi. 

## Contribution Guidelines


BenjaminDannegard (changes requested):
The tutorial looks great! But I think that we can slightly modify the sketch so that we show how to print the values and then show a screenshot of the values being printed so that the sketch part is more of a complete part of the tutorial. If this is easy to do then I think it is worth adding, otherwise keep it as it is.

BenjaminDannegard (approved):
LGTM

PR #1262, opened 2023-08-10
[AE-139] Correct I2C & SPI labelling in MKR Shield `tech-specs.yml` files

aliphys:
## What This PR Changes
- The Docs page ,specifically `tech-specs.yml`, is revised to ensure that
     - The protocol (SPI/I2C/other) is correct
     - The pins used for said protocol are documented (they are not all common)
     - CS pin (if applicable) is defined. (RGB Shield may not need a CS pin?)

## Contribution Guidelines


sebromero (approved):
Looks good to me given that Ali thoroughly dug up this information. We are merging this. @jacobhylen if you find any issue, please report so we can open another PR.

PR #1305, opened 2023-09-05
Include newgrp - docker instead of sudo

martab1994:
## What This PR Changes
- While launching adb shell, we should launch newgrp - docker command instead of sudo
- 
## Contribution Guidelines
- [X] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


jcarolinares (approved):
Double checked with @MaxPayne86 

Both agreed this change makes sense and it is better.

PR #1319, opened 2023-09-08
Update cheat-sheet.md - fix: Table of IO pins

felipenmartins:
fix: Table of IO pins

## What This PR Changes
Change type of pin A0 from "Digital" to "Analog". Change type of pin A3 from "Analog In" to "Analog" to account for the fact that it is also the OpAmp output.

## Contribution Guidelines


jacobhylen (approved):
Hej @felipenmartins.
Nice catch, lgtm. 

Thanks for your contribution! ☀️

PR #1334, opened 2023-09-14
Fix: Arduino Nano has only 20 digital pins max.

aentinger:
## What This PR Changes
Fix: Arduino Nano has only 20 digital pins max: D0 - D19 (which contains A0 to A5). A6/A7 can only be analog intputs.

## Contribution Guidelines


sebromero (approved):
LGTM

PR #1348, opened 2023-09-19
[MKC-1173] Rework IDE 2 core installation guide

jacobhylen:
## What This PR Changes
- This PR reworks the core installation guide to give a slightly more complete flow per-board instructions. 
- Current approach reworks it in a way that does not include screenshots and would therefore be easier to maintain, whether this approach is the correct one is up for discussion. 

## Contribution Guidelines


jacobhylen:
Hej @karlsoderby 
I tried an approach here that doesn't have any screenshots for each core package, with the intent to keep maintenance efforts low. 

I think it works, but let me know what you think.

karlsoderby:
@jacobhylen , yes much better without screenshots and saves a lot of maintenance work. A few things to be updated:
- Installation instructions should be at the top. 
- Installation instructions should also include a level 3 header (###) that explains how to install a specific version of the board package as well.
- I feel the board sections are a bit bare right now, and needs some introduction. See my suggestion below:

---
### UNO R4 Boards

The **UNO R4 Boards** package is based on the **Arduino Renesas Core**, which includes the following boards:
- UNO R4 WiFi
- UNO R4 Minima

You can install this package in the board manager by searching for **"Arduino UNO R4 Boards"**. For detailed instructions, see the [Installing a Board Package](#installing-a-core) section. 

For source code and reporting issues, please visit the official GitHub repository:
- [ArduinoCore-renesas](https://github.com/arduino/ArduinoCore-renesas)

---

jacobhylen:
@karlsoderby Good feedback, I pushed changes that address your comment. Wdyt?

karlsoderby (approved):

PR #1424, opened 2023-10-17
Bug-fix: cd into mounted volume in docker container [2]

MaxPayne86:


jcarolinares (approved):
Perfect

Thank you @MaxPayne86 about the resolution of this bug!

Merged

PR #1380, opened 2023-10-02
[AE-138][MKC-1206] Correct/improve MKR Connector Graphic

aliphys:
## What This PR Changes
- This PR introduces a new MKR Connector graphic, developed in alignment with @facchinm (HW) and Fabio Ferroro (Design)
- The number of covered boards is expanded from 10 to 14
- `RESET`, I2C/SPI pins are colored in line with the colour schema presented by Design
- A note is added, regarding the concurrent use of multiple MKR shields with over lapping CS pins. This should reduce the confusion by users when using multiple shield together
    - Example: https://www.reddit.com/r/arduino/comments/d2vg8r/can_multiple_therm_shields_be_used/
- The `pinout.png` file was not present for the IMU, Relay Proto or SD Proto shields. This is now added in this PR.

bild

## Contribution Guidelines
- [ x] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.

PR #1426, opened 2023-10-18
Fix: In MKR Relay Shield `AREF` and `A0` are mistakenly noted as `SDA` and `SCL` respectively.

aliphys:
## What This PR Changes
- In the https://github.com/arduino/docs-content/pull/1380 , for the MKR Relay Shield `AREF` and `A0` are mistakenly noted as `SDA` and `SCL` respectively.
- This PR removes this label from the pin. `SDA` and `SCL` are on pins `D11` and `D12` respectively. This is not modified.
image

## Contribution Guidelines


karlsoderby (changes requested):
Hey @aliphys , looks great but a very very small design change is needed, which are the rounded corners when it is an ending label.

I already made the change in Figma, but can you export and update the images?

karlsoderby (approved):

PR #1487, opened 2023-11-09
DTB section updated

jcarolinares:
## What This PR Changes
- Updates regarding the DTB

## Contribution Guidelines

### DIFF
--- a/content/hardware/04.pro/carriers/portenta-hat-carrier/tutorials/user-manual/content.md
+++ b/content/hardware/04.pro/carriers/portenta-hat-carrier/tutorials/user-manual/content.md
@@ -2621,7 +2621,7 @@ During the early boot stage, these overlays are merged together into a single DT
 
 ### Handling DTB Overlays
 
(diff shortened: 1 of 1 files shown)


jcarolinares:
cc @TaddyHC we are updating this section, @MaxPayne86 identified that the first method is not available anymore for a long time so I modified the section

cc @martab1994 thank you all for your collaboration to make this great!

MaxPayne86:
@jcarolinares 

> a couple of methods
> In the case of the Portenta X8 with the Portenta Hat Carrier, the following official methods are supported

need to be amended too, the first because we have only one method now which is setting overlays u-boot variable. The second because this mechanism works always and is not tied to a particular carrier. More precisely the overlays are the hw configuration that the X8 is going to apply to itself, in order to support a certain hw

TaddyHC:
@jcarolinares okay, thanks for the notice!

jcarolinares:
@MaxPayne86 changed it to a more generic phrase that hopefully will work!

MaxPayne86 (approved):

PR #1501, opened 2023-11-16
Fix voltage calculation - Update mkr-battery-app-note.md

fl43:
I was wondering why the calculated voltages in the example have not been correct. So I analysed it and found a wrong data type (max_Source_voltage).

## What This PR Changes
- Fix data type of variable max_Source_voltage. Float is correct here to get correct values.
-  Made calculation of the two voltages (voltADC and voltBat) better. No real change, but more accurate results this way. The order of the operations matter when you want to get accurate results.

## Contribution Guidelines

### DIFF
--- a/content/hardware/01.mkr/01.boards/mkr-wifi-1010/tutorials/mkr-battery-app-note/mkr-battery-app-note.md
+++ b/content/hardware/01.mkr/01.boards/mkr-wifi-1010/tutorials/mkr-battery-app-note/mkr-battery-app-note.md
@@ -124,7 +124,7 @@ We will go through the lines needed to create a Sketch to read the battery value
 **4.** We will now create a variable to store the maximum source voltage `max_Source_voltage` as well as the upper (`batteryFullVoltage`) and lower (`batteryEmptyVoltage`) values for the battery. We will also define the battery capacity as `batteryCapacity` so as to determine the charging current. Since we are using a 750 mAh battery in this example, we will set the value to `0.750`.
 
 ```arduino
-  int max_Source_voltage;
+  float max_Source_voltage;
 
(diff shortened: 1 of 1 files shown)


per1234 (approved):
I verified that the previous `max_Source_voltage` type was inappropriate and resulted in incorrect readings due to integer truncation. I verified that the proposed fix to the bug is correct.

I also verified that the other refactorings proposed here are valid.

Thanks @fl43!

content/hardware/01.mkr/01.boards/mkr-wifi-1010/tutorials/mkr-battery-app-note/mkr-battery-app-note.md
```diff
@@ -185,12 +185,12 @@ void loop()
 **13.** In order to convert `rawADC` into a voltage reading (`voltADC`) we will divide `rawADC` by 4095 and then multiply it by the analog reference voltage (3.3V). 
 
 ```arduino
-voltADC = rawADC * (3.3/4095.0);
+voltADC = rawADC * 3.3 / 4096.0;
 ```
 
 **14.** The `voltADC` variable gives us the voltage sensed directly on the PB09 pin. This voltage is passed through the voltage divider, so it is a fraction of the actual battery voltage. We can then calculate the equivilanet battery voltage as follows.
 ```arduino
```
jhansson-ard:
```suggestion
**14.** The `voltADC` variable gives us the voltage sensed directly on the PB09 pin. This voltage is passed through the voltage divider, so it is a fraction of the actual battery voltage. We can then calculate the equivalent battery voltage as follows.

```arduino
```

PR #1513, opened 2023-11-20
[MKC-1331] MicroPython 101 Tweaks

jacobhylen:
## What This PR Changes
This PR fixes a couple oversights in the MicroPython 101 Course:
- When we've reordered tutorials within the course, we didn't change the "next chapter" links, so if you follow them the order is jumbled and incorrect.
- In the installation guide, we were listing the tools to be installed, with links. The links may seem initially like they would be a good idea, but they inherently prompt readers to click, and download. Instead, we should prefer them to wait with downloading until they get to the step by step guide later on the page, for a more intuitive flow. 
- One empty link.  

## Contribution Guidelines

### DIFF
--- a/content/micropython-course/course/01.installation/installation-tools.md
+++ b/content/micropython-course/course/01.installation/installation-tools.md
@@ -16,8 +16,8 @@ In this chapter we will go over how to install the prerequisite software: the **
 
 The software you will have installed are:
 
-- [Arduino Lab for MicroPython](https://labs.arduino.cc/en/labs/micropython)
-- The [Arduino MicroPython Installer](https://labs.arduino.cc/en/labs/micropython-installer)
+- Arduino Lab for MicroPython
+- The Arduino MicroPython Installer
 
 ![The Installed Software](./assets/apps-open.png)
 
@@ -39,7 +39,7 @@ The **Firmware Installer** program will help install MicroPython on your board.
 
 It will automatically download the latest version of the MicroPython firmware and install it on your board, so that all you need to do is to connect the board and press a button.
 
-First [download the app here](), and extract the files to a folder on your computer.
+First [download the app here](https://labs.arduino.cc/en/labs/micropython-installer), and extract the files to a folder on your computer.
 
 If you're on macOS, move the application to your **Applications** folder.
 
--- a/content/micropython-course/course/01.installation/installation-tools.md
+++ b/content/micropython-course/course/01.installation/installation-tools.md
@@ -7,17 +7,17 @@ title: '2. MicroPython Installation Guide'
 description: 'Learn how to install a code editor needed to program your board with MicroPython.'
 ---
 
-## Goals
+## Software Required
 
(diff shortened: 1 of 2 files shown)


content/micropython-course/course/01.installation/installation-tools.md
```diff
@@ -16,8 +16,8 @@ In this chapter we will go over how to install the prerequisite software: the **
 
 The software you will have installed are:
 
-- [Arduino Lab for MicroPython](https://labs.arduino.cc/en/labs/micropython)
-- The [Arduino MicroPython Installer](https://labs.arduino.cc/en/labs/micropython-installer)
+- Arduino Lab for MicroPython
```
karlsoderby, line 19: I would not remove the links here. I know they are listed further down, but I don't think it hurts having them here as well (some people might just want a quick access to the download link).

Maybe we can instead create a section like this at the top:
```md
## Software Requirements
In this chapter, we will be using the following software:
- [Installer]()
- [Editor]()
```

karlsoderby (commented):
I think while we are at it, let's consider changing the top part of this tutorial. It starts with `## Goals` and then goes directly in to it. There's currently no other chapter having a "Goals" section as well.

I think maybe it would best to have a gentle introduction at the very start, along the lines of:

```md
To use MicroPython on an Arduino, we need the two following tools:
- [MicroPython Firmware Installer]() - a tool that installs a specific firmware on your board, so that you can program it using MicroPython
- [MicroPython Lab for Arduino]() - an editor that you can program your Arduino in (like the Arduino IDE).
```

Doesn't need to be exactly like ^, but some sort of intro!

content/micropython-course/course/01.installation/installation-tools.md
jacobhylen: The idea came after Giulio Tamberi went through the course and got confused and jumbled by exactly this, so it can definitely hurt the experience.
The crux is that seeing the link will tempt even the people who don't just want a quick link to leave the microPython 101 page for something else, and fall out of the courses linear path there. 

I do see your point about someone who might just want to find quick access to links though. 

I think we can find a middle ground. :)

jacobhylen:
@karlsoderby How's this? :)

content/micropython-course/course/01.installation/installation-tools.md
```diff
@@ -7,17 +7,17 @@ title: '2. MicroPython Installation Guide'
 description: 'Learn how to install a code editor needed to program your board with MicroPython.'
 ---
 
-## Goals
+## Software Required
 
 In this chapter we will go over how to install the prerequisite software: the **code editor** and the **MicroPython firmware installer**, that will enable you to dive into the learning activities of this course.
```
karlsoderby, line 12:
```suggestion

In this chapter we will go over how to install the prerequisite software: the **MicroPython firmware installer** and the **Arduino Labs for MicroPython (Code Editor)**. The installer will install a MicroPython firmware on your board, and the code editor will allow you to program the board with MicroPython.

## Software Required
```

Looks great @jacobhylen , I am suggesting placing the paragraph _before_ the `## Software Requirements`.

PR #1613, opened 2023-12-15
Add missing temperature variable

alranel:
The temperature variable is missing from most screenshots of this tutorial as well as the initial list, but is present in the code example thus if the user doesn't create it in the cloud the sketch won't compile. This PR adds the variable to the list of variables to create - the screenshot will still lack it, but this fix is better than nothing. :)


karlsoderby (approved):
Thanks for updating it @alranel. Yes and we can later on add a new screenshot! LGTM

PR #1658, opened 2024-01-13
Fix computation of half wavelength in I2s-library-examples.md

Narcolessico:
Dividing the sampleRate by the frequency results in the number of samples per complete wave. For example: with a sampling rate of 24 and a 4Hz frequency we have 24/4 -> 6 samples, which represent a wave peak to peak; we still need to divide by 2 to have its half.

## What This PR Changes
The PR modifies an example code, fixing what I believe is a small mistake. I am not sure if the same mistake is also anywhere else.

## Contribution Guidelines


Hannes7eicher (approved):
Hello @Narcolessico. 

Good catch and thanks for the contribution!

I'll merge it right away.

PR #1681, opened 2024-01-19
Update shield-compatibility.md

Hannes7eicher:
## What This PR Changes
This PR updates the shield compatibility article for the UNO R4. The NineAxesMotion needs an update to work with the UNO R4.

## Contribution Guidelines


karlsoderby (changes requested):
@Hannes7eicher I think it is actually better to remove it, as the above sentence states which shields are compatible, not the ones who are not.

karlsoderby (approved):
LGTM!

PR #1716, opened 2024-01-29
Upgrade datasheets action

Matteo-it:
Upgrade datasheets action and node version, since we have some problem on rendering: https://github.com/arduino/docs-content/actions/runs/7695157420/job/20967436051


sebromero (approved):
LGTM

PR #1738, opened 2024-02-06
[Arduino Cloud] Cloud Agent Mini-article

karlsoderby:
## What This PR Changes
- Very brief article explaining the cloud agent, and how to install it

## Contribution Guidelines


jhansson-ard:
Just a thought, maybe we should link to this HC article explaining the download process as well? 
https://support.arduino.cc/hc/en-us/articles/360014869820-Install-the-Arduino-Create-Agent

cc @seaxwi

karlsoderby:
Sure thing, I added the link at the end of the article in a section called troubleshooting.

jacobhylen:
I'm not sure if I think this is the right approach.
I believe there is a need for this information on docs, but I don't like the idea of having the same/very similar content in two places. It will be difficult to maintain and confusing for users.

Maybe we can move the article from HC to docs, or create a dummy article on docs that we redirect to the HC one so there is an entry point from docs?

jhansson-ard:
I agree, the information should be in one place - not duplicated. Did you check with @seaxwi when writing the article? Perhaps he has some ideas as well.

For the "entry point", where do we intend users to face this issue most? Perhaps we can link the HC article on the Cloud page?

karlsoderby:
@jacobhylen @jhansson-ard definitely best in one place, however in this case I don't mind it existing in two places because it is only a couple of paragraphs, and the HC article needs to set the context for what the Cloud Agent is.

Either way, I need to add this to the docs as it is a very important element of the cloud, and for the flow of the documentation. So I suggest let's go ahead and merge and then discuss with @seaxwi later if we want to update the HC article.

PR #1807, opened 2024-03-11
[PC-1665] Full pinout added to H7 Lite and Connected versions

jcarolinares:
## What This PR Changes
- @jhansson-ard and @sebromero noticed there are not full pinouts for the lite and connected versions of the Portenta H7
- This PR put them in case we consider makes sense
- However **let's keep in mind all the boards share the pinout but not all the boards have all the modules, which can create confusion in some final users**
- Therefore @martab1994 please tell me if you want to do it or not, and I will merge or archive the PR

Thanks

## Contribution Guidelines

### DIFF
--- a/content/hardware/04.pro/boards/portenta-h7-lite-connected/downloads/ABX00042-full-pinout.pdf
+++ b/content/hardware/04.pro/boards/portenta-h7-lite-connected/downloads/ABX00042-full-pinout.pdf
@@ -1,0 +1,1 @@
+../../../../04.pro/boards/portenta--h7/downloads/ABX00042-full-pinout.pdf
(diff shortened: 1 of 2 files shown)


content/hardware/04.pro/boards/portenta-h7-lite-connected/downloads/ABX00042-full-pinout.pdf
```diff
@@ -0,0 +1 @@
+../../../../04.pro/boards/portenta--h7/downloads/ABX00042-full-pinout.pdf
```
jhansson-ard, line 1: These are not shown in the preview. I'm not sure we can symlink downloadables like this, are we doing it for another product where I can check?

jcarolinares: to check in the future, it works for tutorials but I guess the backed is not prepared for "downloadables"

good point indeed , thanks

PR #1907, opened 2024-04-29
Removing validation on main

lukdog:
## What This PR Changes
We 're spending a lot of minutes of GH action also slowing down other jobs for content validation on main.
It is not really important because the validation is already performed on PRs, and the content can not be merged if the validation in PR fails.

## Contribution Guidelines

### DIFF
--- a/.github/workflows/workflow-validate.yaml
+++ b/.github/workflows/workflow-validate.yaml
@@ -3,9 +3,6 @@ name: Validate Content
 on:
   repository_dispatch:
     types: deploy
-  push:
-    branches:
-      - main
   pull_request:
     types: [opened, synchronize, ready_for_review, reopened]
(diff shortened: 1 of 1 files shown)


lukdog:
> @lukdog does this mean that the validations won't run on a "push to production" action but we'll still have them on PRs?

At the moment the Validation was not running on deploy to production, but only after a merge to main branch.
I think it is not really needed because the PRs are already checked, and no-one is checking the status of this validation after merge.

But maybe, a good idea can be to apply this, but add the validation on deploy, blocking non valid content.
WDYT?

jhansson-ard:
Yes, then we can remove this - like you said, the PRs should already be checked. If the validation breaks there we should not be able to push to main :)

PR #1970, opened 2024-05-22
small typo

g-castronovo:
## What This PR Changes

## Contribution Guidelines


jcarolinares (changes requested):
hey @TaddyHC can you double check this change?

image


I want to make sure it is indeed a typo

TaddyHC:
@jcarolinares I have just checked with the PLC IDE and it is typo which I made. I may have written as "IO Type" to separate and give better idea that it's about Input/Output Type but the PLC IDE is showing as IO Type. So to follow consistency and so on, the correct expression for that case is "IOType". 

Also, if you guys would like, since we have this [PR](https://github.com/arduino/docs-content-private/pull/668/files) for the same content with major updates that is about to be published. So we could migrate the changes from present PR to the other one, have one PR that makes all the changes at once, and just to avoid version conflicts. What do you say?

g-castronovo:
Sure, please go ahead and close this one then!

PR #2020, opened 2024-06-14
[PC-1707] - Add I2S Interface to the Portenta's C33 Tech Specs Table

canchebagur:
## What This PR Changes
- This PR adds the I2S interface to the Portenta's C33 tech specs table. 

## Contribution Guidelines


jcarolinares (approved):
Double checked with shematics

Changes are correct, I2S is only accessible using the High density connectors which means that a carrier will be need it to illustrate the examples 

merging cc @martab1994 @canchebagur

PR #2091, opened 2024-07-30
[MKC-1756] Removes old cloud editor tutorial

jacobhylen:
## What This PR Changes
- Currently there are two similar tutorials on the cloud docs, one for the old web editor that doesn't exist anymore, and one that is labelled "(new)" that features the new cloud editor that is currently the one that is available. The old article needs to be removed. 

A redirect will be set up so we don't break links to the old tutorial.

## Contribution Guidelines


g-castronovo:
@jacobhylen the only thing I do not like about the new tutorial is that we removed the screenshot /part of the Agent pop up wizard. Do you think we could add it?

jacobhylen:
Sure, I'll take a new screenshot and add it

g-castronovo (approved):
<*)))]><

PR #2098, opened 2024-08-01
11885 - Added note for 3.3V boards (ReadAnalogVoltage)

pedromsousalima:
## What This PR Changes
- Added note for 3.3V boards

## Contribution Guidelines


jacobhylen (changes requested):
We should also consider that we have board with different ADC resolutions.
Can we add a comment about this too?

pedromsousalima:
@jacobhylen Good point. Added information on that too! Thank you

jacobhylen (approved):
LGTM

PR #2104, opened 2024-08-05
Update esp-now document

buhuihongshide-sdjjd:
## What This PR Changes
the callback function of receiving the ESPNOW data is "esp_err_t esp_now_register_recv_cb(esp_now_recv_cb_t cb)",and the define of esp_now_recv_cb_t is "typedef void (*esp_now_recv_cb_t)(const esp_now_recv_info_t *esp_now_info, const uint8_t *data, int data_len)"which is not"typedef void (*esp_now_recv_cb_t)(const uint8_t *macAddress, const uint8_t *data, int data_len)"

## Contribution Guidelines

### DIFF
--- a/content/hardware/03.nano/boards/nano-esp32/tutorials/esp-now/esp-now.md
+++ b/content/hardware/03.nano/boards/nano-esp32/tutorials/esp-now/esp-now.md
@@ -235,7 +235,7 @@ typedef struct struct_message {
 struct_message myData;
 
 // callback function that will be executed when data is received
-void OnDataRecv(const uint8_t * mac, const uint8_t *incomingData, int len) {
+void OnDataRecv(const esp_now_recv_info_t *info, const uint8_t *incomingData, int len) {
   memcpy(&myData, incomingData, sizeof(myData));
   Serial.print("Bytes received: ");
   Serial.println(len);


jcarolinares:
Hello @Hannes7eicher when you have a moment please take a look at this pull request, thanks!

Hannes7eicher:
Thanks for the PR @buhuihongshide-sdjjd! I've had a quick look and it looks alright to me. But if you try to compile the sketch you will see that `'esp_now_recv_info_t' does not name a type;`, so I think the Arduino Core uses an older version of the ESP-NOW library. I will bring this up with the firmware team and see if we can update it.

buhuihongshide-sdjjd:
> Thanks for the PR @buhuihongshide-sdjjd! I've had a quick look and it looks alright to me. But if you try to compile the sketch you will see that , so I think the Arduino Core uses an older version of the ESP-NOW library. I will bring this up with the firmware team and see if we can update it.`'esp_now_recv_info_t' does not name a type;`

Sorry that I didn't do well test on this code.

Hannes7eicher:
I had a chat with our engineering team and as you correctly pointed out, the signature for `esp_now_recv_cb_t` has indeed changed between different versions of the ESP-IDF. I have summarized the changes in case others have the same concern.

**Current Implementation:**

According to our currently supported version of the Arduino ESP32 Boards core, which is based on the 2.x branch of the arduino-esp32 core and uses ESP-IDF v5.1, the correct callback function signature for `esp_now_recv_cb_t` is:

```c
typedef void (*esp_now_recv_cb_t)(const uint8_t *macAddress, const uint8_t *data, int data_len);
```

This is consistent with the [documentation for version ESP-IDF v5.1.4](https://github.com/espressif/esp-idf/blob/v5.1.4/docs/en/api-reference/network/espnow.rst).
It's important to note that from ESP-IDF v5.1 onwards, Espressif has introduced some changes, including how callback functions are defined. In the newer versions, specifically for the 3.x branch, the callback function signature has been updated to:

```c
typedef void (*esp_now_recv_cb_t)(const esp_now_recv_info_t *esp_now_info, const uint8_t *data, int data_len);
```

**What This Means for Our Core:**

Since our current core still depends on the 2.x branch, the existing callback signature with `macAddress` is valid and correct. The modification you mentioned—using `const esp_now_recv_info_t *esp_now_info`—applies to the 3.x branch of ESP-IDF and is part of the updated API that Espressif has introduced.

For our users who are working with the Nano ESP32 and our supported core version, there is no need to patch or change the existing callback signature. The migration to the newer 3.x API involves several changes, as highlighted in Espressif's extensive migration guide. We advise sticking to the functionality provided by our current core unless you are planning to migrate your entire project to a newer version of the ESP-IDF.

I hope this clarifies your concern. If you have any more questions or need further assistance, feel free to ask! Otherwise I will go ahead and close the issue.

PR #2241, opened 2024-10-14
fix typo, it said sawtooth, it should say sine

montoyamoraga:
later in the tutorial the sawtooth wave is used.

## What This PR Changes
- it changes a mention of sawtooth wave, for sine wave, which is what it meant, it was a typo.

## Contribution Guidelines
- [ X] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


g-castronovo:
@jacobhylen  since you're the original author, I need to ask you: I am not sure is a typo. what kind of curve shows the progenerated code?

jacobhylen:
Hello @g-castronovo 
@montoyamoraga is correct, there was indeed a typo introduced at some point. 
We should also change the comment in the code example [here](https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/AnalogWave/examples/SineWave/SineWave.ino) that also says sawtooth.

karlsoderby (approved):
Thanks for correcting this typo @montoyamoraga , looks good to me.

Issue #2261, opened 2024-10-28
Schematic Issue - Multiple Serial Ports on the Arduino Mega


jglatts:
It looks like the [Use Multiple Serial Ports on the Arduino Mega](https://docs.arduino.cc/built-in-examples/communication/MultiSerialMega/) page has a small error in the schematic example. 

The TX pin of the mega is going to the TX pin of the Serial-enabled device, as opposed to the RX pin of the Serial device. 
The RX pin of the mega is also going to the RX pin of the Serial device. 

The deleted asset in commit 4aed0c17a67c5b5ce24cf624c5492bd7baf64a35 appears to be the right wiring.

PR #2275, opened 2024-11-10
Fix the MultiSerialMega Example Schematic Image

jglatts:
## What This PR Changes
- Fixes the schematic image on the [MultiSerialMega](https://docs.arduino.cc/built-in-examples/communication/MultiSerialMega/#schematic) example page, 
- #2261 

## Contribution Guidelines
- [✓ ] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


jglatts:
Looking more at the image, the deleted asset is using Serial0 pins which is tied to the monitor. 
The other image is using Serial1 pins, which the example code needs. 
However the Serial1 pins to the UART enabled device is incorrect, will close this PR for now.

PR #2340, opened 2024-12-19
[PXCT-328] OPAMP Library API

pedromsousalima:
## What This PR Changes
- Add reference to OPAMP library

## Contribution Guidelines


Hannes7eicher (commented):
LGTM! What are your thoughts on merging this with the existing OPAMP documentation? They are both relatively short so I think it might be better to have one slightly longer document than two shorter ones.

PR #2459, opened 2025-04-20
[PXCT-899] Arduino UNO R4 - Update led-matrix.md

joshuajka:
we are setting the pixel value from the third row and second column in this case frame[2][1] , and not how was writted before

## What This PR Changes
- changes the text explaining the example of the function renderBitmap(frame, 8, 12);

## Contribution Guidelines
- [ x ] I confirm that I have read the [contribution guidelines](https://github.com/arduino/docs-content/tree/main/contribution-templates) and comply with them.


jcarolinares:
Merging, thanks again @joshuajka for the fix!

PR #2488, opened 2025-05-16
Downgrade ubuntu, add sudo to ds renderer

Matteo-it:


.github/workflows/preview.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   run:
     if: ${{ contains(github.event.pull_request.labels.*.name, 'preview') || github.ref_name == 'main' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
lukdog, line 17: is this really needed?

Matteo-it: Ok ds-rendering action is passed, restored

lukdog (approved):

PR #2494, opened 2025-05-20
Fix board-registration entrypoints links

Matteo-it:
Fix entrypoints links for new board-registration flow


jhansson-ard (approved):
LGTM!

PR #2528, opened 2025-06-06
Update eeprom.md

FPA-LowellTroyer:
## What This PR Changes
Changed values in both examples to correct their functionality

Example 1:
val -> value (no longer gives error)

Example2
addr -> value (prints the correct variable instead of the address)

## Contribution Guidelines


jcarolinares (approved):
Thank you!

cc @karlsoderby user is right

PR #2530, opened 2025-06-11
[PXCT-969] add tiny-ml-kit step file

Hannes7eicher:
## What This PR Changes
This PR adds the step file to the tiny ml kit

## Contribution Guidelines


TaddyHC (approved):
@Hannes7eicher ! I agree with @jhansson-ard 's point in this case, since the hardware is in unlisted directory and so on. I'm not sure if the file should appear within the downloadable resources in this case (I think it should still appear, even if it is in unlisted directory), so I'd like to suggest by compressing the step file as we do and name as AKX00028-shield-step for example. Otherwise, based on the observation until now, lgtm

PR #2557, opened 2025-07-01
[PXCT-1052] Add Modulino MicroPython library to essentials lists 

pedromsousalima:
Added the Modulino MicroPython Package as a suggested library for all Modulino node essentials pages. Also updated the Arduino Modulino library name for consistency.

## Contribution Guidelines

### DIFF
--- a/content/hardware/11.accessories/modulino-nodes/modulino-buttons/essentials.md
+++ b/content/hardware/11.accessories/modulino-nodes/modulino-buttons/essentials.md
@@ -1,8 +1,11 @@
-
 <EssentialsColumn title="Suggested Libraries">
 
-<EssentialElement title="Arduino Modulino" type="library" link="https://www.arduino.cc/reference/en/libraries/modulino/">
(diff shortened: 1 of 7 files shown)


content/hardware/11.accessories/modulino-nodes/modulino-buttons/essentials.md
jhansson-ard, line 1: Do we want to add this section instead?

image

image

pedromsousalima: @jhansson-ard fair. This was my first thought: when doing this, though the library gets highlighted, and that is great, the Micropython package one will not. The fact that one has the Arduino tag and the other does not looks, in my opinion, a bit weird, as they are both official. I can make these changes, but we can also alternatively add it without the map for now, find a way to support micropython packages too, and then add the map for both to ensure consistency across both libraries at all times. Let me know what you think is the best approach :)

jhansson-ard: Thank you for the explanation! We'll do it your way, I trust your judgement 🙏🏼

jhansson-ard (approved):

PR #2619, opened 2025-09-04
added 'beautified' datasheet

jhansson-ard:
## What This PR Changes
Adding the MKR WAN 1310 datasheet based on an older version.

## Contribution Guidelines


jhansson-ard:
@canchebagur this is the PR, I will share the original PDF with you so that you can double check it 🙏🏼

PR #2620, opened 2025-09-04
Fix incorrect VCC voltage for nano connector carrier

sebromero:
Updated VCC voltage specification in the connection table.

## What This PR Changes
- Changes the voltage on the power pin of the Qwiic connector to fixed 3V3. It's always translated to 3V3.

## Contribution Guidelines


sebromero:
Has apparently been fixed in the mean time

PR #2633, opened 2025-09-17
Fix the example

mzaffalon:
-  bugfix: the method WiFiServer::write was removed in https://github.com/arduino/ArduinoCore-mbed/commit/18b696b224f8c621cd95fb099e63e25c68245c96

- modify the example keep the connection open until the client quit

## Contribution Guidelines


mzaffalon:
I am new to Arduino and I didn't follow the development. It appears that there are two WiFi libraries, the other one at https://github.com/arduino-libraries/WiFi which is retired as the note at the beginning of the page says, but this is still referred to in the WiFi library reference.
So is it a matter of updating the documentation?

PR #2711, opened 2025-11-09
Change the order of the pins in the constructor

DrakeTheCoder:
Since Stepper.h library uses `1010`, `0110`, `0101`, `1001` sequence for four-wire connection.

## What This PR Changes
Since Stepper.h library uses `1010`, `0110`, `0101`, `1001` sequence for four-wire connection shouldn't 2th and 3th PINs being swapped places inside the constructor? For all examples.

https://github.com/arduino-libraries/Stepper/blob/a62defa855f312111028fec0bfab38ab64b4ddcc/src/Stepper.cpp#L256

## Contribution Guidelines


DrakeTheCoder:
I was wrong

PR #2944, opened 2026-05-07
[PXCT-1779] brick docs rework

swikstrom:
## What This PR Changes
- expand and restructure bricks documentation

## Todo
- [x] Verify that the custom brick example code blocks reflect the downloadable archives (`https://content.arduino.cc/assets/simple-web-server-brick.zip` and `https://content.arduino.cc/assets/simple-device-access-brick.zip`)

## Contribution Guidelines


content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self):
+        print("[CustomBrick] Initialized")
+    
+    def do_something(self):
+        print("[CustomBrick] Doing something useful...")
+```
+
+### Docker Containers (`brick_compose.yaml`)
+
+If your Custom Brick requires external services, you can define them using a Docker Compose file. The system will automatically pull and run these containers alongside your App.
```
per1234:
```suggestion
If your Custom Brick requires external services, you can define them in this file. `brick_compose.yaml` is a Docker Compose file:

https://docs.docker.com/reference/compose-file/

The system will automatically pull and run these containers alongside your App.
```

- Adjust wording to make it clear that `brick_compose.yaml` is the referenced Docker Compose file.
- Link to the essential high quality Docker Compose file documentation

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,148 @@
...
+
+```yaml
+services:
+  my_service:
+    image: custom_registry/my_service:latest
```
per1234:
```suggestion
    image: some_registry/some_image:latest
```

- Adjust placeholder reference components to avoid giving the false impression that use of an image created by the user is required. Existing images may also be utilized in custom Bricks.
- Use correct "image" term ([reference](https://docs.docker.com/reference/compose-file/services/#image)).

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,148 @@
...
+    ports:
+      - "8080:8080"
+```
+
+<Alert type="warning">**Important:** Docker images specified in `brick_compose.yaml` must be publicly accessible, as App Lab does not currently support private registries for Custom Bricks.</Alert>
+
+## Using Your Custom Brick
+
+Once created, using a Custom Brick is identical to using a built-in Brick. Import the class from your package and initialize it in your App's `main.py`.
```
per1234:
```suggestion
Once created, using a Custom Brick is identical to using a built-in Brick. Import the class from the Python package in the Brick and initialize it in your App's `main.py`.
```

Adjust wording to make it more clear what is being referred by "your package".

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,148 @@
...
+my_brick = CustomBrick()
+
+# Use its functionality
+my_brick.do_something()
+
+# Start the App
+App.run()
+```
+
+## AI Models in Custom Bricks
```
per1234, line 248: This is certainly better than no documentation of the feature at all, so I'm fine with shipping the initial version of the docs with this as-is. However, I feel that it doesn't provide nearly enough information for the average user to be able to utilize the feature without reverse engineering how it is done in the built-in Bricks.

per1234 (changes requested):

swikstrom:
Sorry @per1234 , I completely overlooked your comments when closing this PR.

The scope of this task expanded to include a deeper rework of the Bricks documentation (unfortunately, taking longer than expected).

I've reopened the PR and will now address your comments!

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: 3f3e5a73c9f9b7e15bde362bc156cb20f587a6ad adds the link to `custom-bricks.md`, as well as the new `bricks-reference.md` article.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: I've applied the feedback in 466cedd7538e36a8b8d4a452a42e7d77988f57e4

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: See d524f4bee31aae744f6b921cae1c3c0e8ae9229b

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: @per1234 In addition to the example with placeholder components, do you think we should include some real examples?

For example, the one you shared [here](https://forum.arduino.cc/t/uno-q-exploring-custom-bricks-and-app-lab-service-containers/1442898/2):

```yaml
services:
  player:
    image: debian:bookworm-slim
    command: sh -c "ls -l /dev/snd || true; sleep infinity"
    devices:
      - /dev/snd
```

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: I'm working on adding more information, but as long as the current draft is not misleading, let's ship it and tackle the additions in a separate PR.

content/software/app-lab/5.bricks/1.about-bricks/about-bricks.md
```diff
@@ -1,38 +1,46 @@
...
 
-When you add a Brick to your App via the App Lab interface, the system automates several steps:
+Technically, a **Brick** is a Python module that interacts with your main application. Bricks run in one of two ways:
+
+- **Pure Python Bricks:** These run entirely in the board's main Python environment. They provide utility functions, compute logic, or cloud connections without launching external containers.
+- **Containerized Bricks:** These act as Python bridges to isolated **Docker containerized services** running in the background. Your application script (`main.py`) interacts with the Brick's Python API, which handles any necessary network communication with the underlying container behind the scenes.
+
+While App Lab provides a wide catalog of pre-built Arduino Bricks, you can also create **Custom Bricks** directly within your application as a local variant. Custom Bricks use the exact same architecture as pre-built Bricks, allowing you to package your own specialized Python logic or third-party Docker containers into modular, reusable components.
```
per1234:
```suggestion
While App Lab provides a wide catalog of pre-built Arduino Bricks, you can also create **Custom Bricks** directly within your application. Custom Bricks use the exact same architecture as pre-built Bricks, allowing you to package your own specialized Python logic or third-party Docker containers into modular, reusable components.
```

Remove superfluous "word salad"-y content.

content/software/app-lab/5.bricks/2.use-bricks/use-bricks.md
```diff
@@ -9,55 +9,79 @@ tags:
...
-```python
-web_ui = WebUI()
-```
+After adding a Brick, you must import its Python package into your App's `main.py` file to instantiate its classes or call its functions.
 
-### 3. Launch the App
+### Managed Bricks
 
-You must include the `App.run()` function at the very bottom of your `main.py` file. This command acts as the orchestrator, launching all imported Bricks and starting the Bridge communication.
+The vast majority of Bricks follow a managed execution pattern. You import the class, instantiate it to register it with the orchestrator, and call its methods:
```
per1234:
```suggestion
The vast majority of Bricks utilize the App framework's lifecycle management system. You import the class, instantiate it to register it with the system, and the App framework automatically manages its execution:
```

---

Use the established "lifecycle" term:

https://github.com/arduino/app-bricks-py/blob/0bc3c36b88552c4d9acb6fe0daa083c02f74473f/src/arduino/app_utils/app.py#L17

https://github.com/arduino/app-bricks-py/blob/0bc3c36b88552c4d9acb6fe0daa083c02f74473f/src/arduino/app_utils/brick.py#L35

I had to adjust the wording of the sentence a bit in order to accomplish this, since I believe it is most accurate to refer to "lifecycle" in the larger context of a Brick being used by an App (as opposed to only in a brick-specific context).

---

Avoid the use of the term "orchestrator". It seems there will be a tendency to try to use that term to refer to the App management functions of Arduino App CLI, so unnecessarily using the same term to refer to something completely different would be likely to cause confusion to the reader.

---

Provide a brief explanation of what "lifecycle management" means. The meaning will not be obvious to the average reader from the term "lifecycle management" (or "managed execution") alone.

---

Remove "call its methods". A "managed Brick" won't necessarily have user called methods, so it doesn't make sense to imply this is an inherent characteristic of using managed Bricks.

content/software/app-lab/5.bricks/2.use-bricks/use-bricks.md
```diff
@@ -9,55 +9,79 @@ tags:
...
+
+1. Import the Brick's Python package from the `arduino.app_bricks` namespace (or directly by its folder name if it's a Custom Brick).
+2. Instantiate the Brick class (for class-based Bricks) or call its plain functions directly.
```
per1234:
```suggestion
2. Instantiate the Brick class.
```

The lifecycle management system requires the use of a class by the Brick, so it doesn't make sense to imply this is an optional step of using managed Bricks.

content/software/app-lab/5.bricks/2.use-bricks/use-bricks.md
```diff
@@ -9,55 +9,79 @@ tags:
...
+# 3. Launch the App and all Bricks
 App.run()
 ```
 
-<Alert type="warning">**Warning:** The App framework ignores any code placed after `App.run()`. Ensure all your logic and Brick initializations happen before this call.</Alert>
+<Alert type="note">**Important:** The `App.run()` call typically blocks the main thread. Ensure all your setup logic and Brick initializations happen before this call.</Alert>
+
+### Unmanaged Bricks
+
+Some simpler Custom Bricks or specific official Bricks (like `arduino:streamlit_ui`) don't use managed objects. Instead, you call their functions or import their modules directly, without needing to register them with the App orchestrator.
```
per1234:
```suggestion
Some simpler Custom Bricks or specific official Bricks (like `arduino:streamlit_ui`) don't use the lifecycle management system. Instead, you manage their execution directly from the App's Python script code.
```

- Use the previously established "lifecycle management" term.
- Replace "call their functions". Calling functions or methods is not unique to using unmanaged Bricks. Managed Bricks also often provide user called functions/methods.
- Remove "import their modules directly". Importing modules is not unique to using unmanaged Bricks.
- Remove mention of registration. Registration is handled automagically for the Brick user by the App framework, so this is not something the user considers when using a managed Brick. It is true that the registration requires the user to instantiate the class, but the user will often also need to instantiate a class to use an unmanaged Brick.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+overwriteSidebar: Custom Bricks
+tags:
+  - Bricks
+  - Python
+  - Docker
+  - Custom
+  - Arduino App Lab
+---
+
+**Custom Bricks** let you package your own Python modules or third-party Docker services into reusable components. Build a feature once, package it as a Custom Brick, and reuse it across your projects.
```
per1234:
```suggestion
**Custom Bricks** allow you to create additional Docker services and companion Python packages for use by an App.
```

## Reduce prominence of Python component

At least with the current App-local custom Brick feature, the sole value of the Python component of the custom Brick is to facilitate the use of the service/container. If the user only wants to organize their App code into Python modules, they should just add packages/modules under the `python` folder of the App. Using a custom Brick for this purpose only adds unnecessary complexity for absolutely no benefit.

## Replace "module" with "package"

Although it is true that a custom Brick could, provide modules alone, this is not the "golden path". When a custom Brick is created, it is populated as a Python package. Thus, I think it will be quite rare for a custom Brick developer to delete that package and add bare modules instead. For this reason, "Python package" is more likely to be correct than "Python module" in this context.

## Remove mentions of reusability

The App-local custom Brick feature we have currently is mostly just a way to achieve additional functionality for a single App. It is true that the custom Brick could be reused by manually copying it to another App, then manually editing an entry for that brick into the `app.yaml` file. However, doing so would be more of a "hack" than something we support as part of the "golden path".

So it definitely doesn't make sense to advertise reusability as a prominent feature of custom Bricks. Moreover, I feel it would only make sense to mention the possibility of reuse if detailed instructions were provided for the relatively complex task of achieving such a thing.

It probably is best to just wait until such time as proper support for reusable Bricks is implemented (https://github.com/arduino/arduino-app-cli/issues/207).

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+
+A typical Custom Brick structure looks like this:
+
+```text
+my-app/
+├── app.yaml
+├── python/main.py
+└── bricks/
+    └── my_custom_brick/            # The Brick ID and folder name
+        ├── __init__.py             # Python logic (required)
```
per1234:
```suggestion
        ├── __init__.py             # Python logic (optional)
```

This file is not required. Here is an example App without a Python package:

https://github.com/user-attachments/files/27430584/Custom.Brick.With.Container.zip

content/software/app-lab/5.bricks/2.use-bricks/use-bricks.md
```diff
@@ -9,55 +9,79 @@ tags:
...
-### 1. Import the Brick Class
+<Alert type="note">**Important:** When you use the App Lab UI, the system manages the `app.yaml` file automatically. Manual changes can cause syntax errors that prevent the editor from opening your project.</Alert>
 
-Import the specific Brick class from the `arduino.app_bricks` module. For example, to use the Web UI Brick:
+### Create a Blank Custom Brick
 
-```python
-from arduino.app_bricks.web_ui import WebUI
-```
+If you need to package your own Python logic or Docker containers into a reusable module, you can create a Custom Brick directly within your application.
```
per1234, line 41:
```suggestion
If your App requires the use of additional Docker containers, you can create a Custom Brick directly within your application.
```

## Remove mention of Python component

At least with the current App-local custom Brick feature, the sole value of the Python component of the custom Brick is to facilitate the use of the service/container. If the user only wants to organize their App code into Python modules, they should just add packages/modules under the `python` folder of the App. Using a custom Brick for this purpose only adds unnecessary complexity for absolutely no benefit.

## Remove mention of reusability

The App-local custom Brick feature we have currently is mostly just a way to achieve additional functionality for a single App. It is true that the custom Brick could be reused by manually copying it to another App, then manually editing an entry for that brick into the `app.yaml` file. However, doing so would be more of a "hack" than something we support as part of the "golden path".

So it definitely doesn't make sense to advertise reusability as a prominent feature of custom Bricks. Moreover, I feel it would only make sense to mention the possibility of reuse if detailed instructions were provided for the relatively complex task of achieving such a thing.

It probably is best to just wait until such time as proper support for reusable Bricks is implemented (https://github.com/arduino/arduino-app-cli/issues/207).

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Create a Custom Brick
+
+You can generate the foundational structure for a Custom Brick directly from the App Lab interface.
+
+1. [Open an App](../../apps/manage-apps/#open-an-app).
+2. Click the **Add Brick** button at the top of the **Editor sidebar**.
+   ![Screenshot of the App Editor in Arduino App Lab, highlighting the Bricks section in the left sidebar.](../../assets/app-lab-editor-hl-bricks.png)
+3. At the bottom of the Bricks catalog, select **Create Custom Brick**.
+4. Enter a name (ID) for your Brick. This must be a valid Python package name (e.g., lowercase letters, numbers, and underscores). <!-- TODO: Verify -->
```
per1234:
```suggestion
4. Enter a name for your Brick.
```

The name provided by the user is the human-friendly name for the Brick. A compliant ID is automatically generated from that human-friendly name. For example, I created a custom Brick with name "Foo Bar-baz", and this resulted in an ID of `foo_bar_baz`:

```yaml
id: foo_bar_baz
name: Foo Bar-baz
```

The Brick folder is named after the ID, so this results in a valid Python package.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Configuration (`brick_config.yaml`)
+
+The `brick_config.yaml` file defines the identity of your Brick and the environment variables it accepts from the App Lab UI.
+
+```yaml
+id: my_custom_brick
+name: My Custom Brick
+description: "A description of what my Custom Brick does."
+category: miscellaneous
```
per1234:
```suggestion
```

These keys don't serve any purpose currently, so there is no reason to include them in this snippet.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+
+```python
+# bricks/my_custom_brick/__init__.py
+def say_hello():
+    print("Hello from the custom brick!")
+```
+
+#### Managed Class-based Bricks
+
+If you want your Custom Brick to behave exactly like an official Brick with automatic lifecycle management (background threads, startup/shutdown hooks), you can use the `@brick` decorator.
```
per1234:
```suggestion
If you want your Custom Brick to behave exactly like an official Brick with automatic lifecycle management (background threads, startup/shutdown hooks), use the `@brick` decorator on the class. The App lifecycle management system will automatically call certain methods of the decorated class:
```

- Remove "you can". The optional nature of the use of lifecycle management was already established at the start of the paragraph. If the Brick developer decides they do wish to utilize the lifecycle management feature, the decorator is mandatory, where as "you can" implies it is optional.
- Specify that the decorator is attached to the class.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+        print("[Brick] Running in background...")
+        time.sleep(1)
+```
+
+### Docker Containers (`brick_compose.yaml`)
+
+If your Custom Brick requires external services (such as databases or companion APIs), you can define them in this file. The `brick_compose.yaml` file is a standard [Docker Compose file](https://docs.docker.com/reference/compose-file/). The orchestrator will automatically pull and run these containers alongside your App. For full details on Docker capabilities and networking, see [Bricks Architecture and Configuration Reference](../bricks-reference/).
+
+```yaml
+ # bricks/my_custom_brick/brick_compose.yaml
```
per1234:
```suggestion
# bricks/my_custom_brick/brick_compose.yaml
```

Remove stray leading space.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+services:
+  my_service:
+    image: some_registry/some_image:latest
+ports:
+  - "8080:8080"
+```
+
+<Alert type="warning">**Important:** The orchestrator executes the custom brick's Python code (in `__init__.py`) within the main application's container, **not** inside the custom Docker containers you define here. This means your Python code can't directly access system libraries or files inside `my_database`. Instead, your Python code must communicate with the containerized service over the virtual Docker Compose network using a network API (such as HTTP, WebSockets, or TCP/IP). You can reach the service using the service name defined in your `brick_compose.yaml` (e.g., `my_database`) as the hostname.</Alert>
```
per1234:
```suggestion
<Alert type="warning">**Important:** The orchestrator executes the custom brick's Python code (in `__init__.py`) within the main application's container, **not** inside the custom Docker containers you define here. This means your Python code can't directly access system libraries or files inside the container. Instead, your Python code must communicate with the containerized service over the virtual Docker Compose network using a network API (such as HTTP, WebSockets, or TCP/IP). You can reach the service using the service name defined in your `brick_compose.yaml` (e.g., `my_service`) as the hostname.</Alert>
```

Correct service name.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+  - Docker
+  - Configuration
+---
+
+This document details the underlying execution model, thread lifecycle management, file directory structures, and configuration specifications for both preinstalled official bricks and custom bricks in the Arduino App Lab ecosystem.
+
+## Core Concepts & Lifecycle Management
+
+Bricks follow one of two execution patterns inside the `arduino.app_utils` framework, depending on whether you implement them as plain functions or managed classes.
+
```
per1234, line 18:
```suggestion

<Alert type="warning">**Warning:** Because `App.run()` blocks the main thread, you must perform all your brick initialization, variable reading, and setup logic *before* calling `App.run()`.</Alert>

```

Preserve the sole relevant information from the otherwise off-topic "App.run() Thread Management" section.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+    # This runs repeatedly on the main thread
+    print("Main thread active...")
+    time.sleep(1)
+
+# App.run blocks the main thread and repeatedly executes my_continuous_loop
+App.run(user_loop=my_continuous_loop)
+```
+
+<Alert type="warning">**Warning:** Because `App.run()` blocks the main thread, you must perform all your brick initialization, variable reading, and setup logic *before* calling `App.run()`.</Alert>
+
```
per1234:
```suggestion
```

This information is completely off topic for the Brick reference.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,183 @@
...
+class MyManagedBrick:
+    def start(self):
+        print("[Brick] Started")
+
+    @brick.loop
+    def my_background_task(self):
+        print("[Brick] Running in background...")
+        time.sleep(1)
+```
+
```
per1234, line 94:
```suggestion

For details on how to utilize the lifecycle management features of the App framework in the Python package of your custom App, see [Bricks Architecture and Configuration Reference](../bricks-reference/).

```

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+## Custom Bricks Directory Structure
+
+Custom bricks are stored locally in your application's `bricks/` folder. The folder name must match the custom brick's configuration ID.
+
+```text
+my-app/
+├── app.yaml
+├── python/main.py
+└── bricks/
+    └── my_custom_brick/            # The Brick ID (must match folder name)
```
per1234:
```suggestion
    └── my_custom_brick/            # The Brick ID (must match id key in manifest)
```

Previously, this would be read as the equivalent of saying "the folder name must match the folder name".

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+- **`requires_display`**: When set to `webview`, this indicates that the brick exposes a graphical interface. This signals the App Lab desktop application to automatically open a new window or browser tab directed to the brick's web interface as soon as the app finishes starting.
+- **`requires_services`**: List of core system services (IDs) that must be running for this brick to function (e.g., `arduino:genie` for LLM bricks).
+- **`required_devices`**: Defines the *classes* of hardware devices the brick requires (e.g., `camera`, `microphone`, `speaker`). When the app is started, the `arduino-app-cli` performs startup validation to ensure at least one physical device of that class is currently connected to the board. If not, an error is raised.
+- **`mount_devices_into_container`**: When `true`, hardware devices (like `/dev/video0`) requested in `required_devices` are automatically mounted into the brick's Docker container.
+
+## Application Configuration (`app.yaml`)
+
+Bricks must be declared in the `bricks` list within the application's `app.yaml` descriptor file. The CLI parses this file to configure the orchestrator.
```
per1234:
```suggestion
Bricks must be declared in the `bricks` list within the application's `app.yaml` descriptor file.
```

The removed sentence doesn't provide anything of value to the reader.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+
+Arduino App Lab automatically manages this section when using the visual interface.
+
+```yaml
+name: "My App Name"
+description: "A description of the app"
+icon: "😀"                 # Must be a single valid emoji
```
per1234:
```suggestion
icon: "😀"                 # Must be a single emoji
```

Unless there is a specific restriction on which emoji may be used, any emoji will be "valid". If there is a restriction (which I doubt), then the specifics must be documented.

content/software/app-lab/5.bricks/1.about-bricks/about-bricks.md
```diff
@@ -1,38 +1,46 @@
...
-2. **Deployment:** The system downloads and starts the required container image on your board.
-3. **Integration:** A dedicated Python library is made available, allowing you to `import` the Brick's functionality directly into your script.
+## How Bricks Work (Orchestration and Infrastructure)
 
-<Alert type="info">**Note:** Most Bricks require an active internet connection the first time they are deployed so the board can download the necessary files from the Arduino registry.</Alert>
+When you add a Brick to your App via the App Lab interface, the system automates several steps:
+
+1. **Configuration:** The orchestrator registers the Brick in your application's `app.yaml` file.
+2. **Environment Preparation:** For containerized Bricks, the `arduino-app-lab` system manages the deployment, virtual network creation, and startup of all required background containers.
+3. **Python Path Integration:** The orchestrator automatically adds the `bricks/` directory to the Python `sys.path`, allowing you to import your Custom Bricks seamlessly by their folder name.
```
per1234:
```suggestion
1. **Configuration:** App Lab registers the Brick in your application's `app.yaml` file.
2. **Environment Preparation:** For containerized Bricks, the `arduino-app-lab` system manages the deployment, virtual network creation, and startup of all required background containers.
3. **Python Path Integration:** App Lab automatically adds the `bricks/` directory to the Python `sys.path`, allowing you to import your Custom Bricks seamlessly by their folder name.
```

I don't see any reason to introduce the term "orchestrator". The reader won't have any familiarity with this term. Conversely, they will easily understand what is written here if "App Lab" is used instead.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+  # Custom brick with configuration variables
+  - my_custom_brick:
+      variables:
+        NAME: "Arduino"
+      model: "my_custom_model_id"  # Optional: Bind a specific AI model
+```
+
+## Containerization and Networking Constraints
+
+Bricks can execute within an isolated background Docker container. You configure and manage this behavior using the following reference-grade rules:
```
per1234:
```suggestion
Bricks can execute within an isolated background Docker container. You can configure and manage this behavior:
```

Remove word salad.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+
+- **Service Dependencies (`requires_services`):** If your brick relies on core containerized system services (such as local AI inference engines), list their system IDs here (e.g., `arduino:genie`). The orchestrator ensures these system services are active before launching your brick.
+- **Hardware Access (`mount_devices_into_container`):** Setting this boolean field to `true` instructs the orchestrator to automatically mount any physical hardware devices matching the classes in `required_devices` (e.g., `/dev/video0`) into the brick's Docker container.
```
per1234:
```suggestion
- **Service Dependencies:** If your brick relies on core containerized system services (such as local AI inference engines), list their system IDs (e.g., `arduino:genie`) in the `requires_services` key of `brick_config.yaml`. The orchestrator ensures these system services are active before launching your brick.
- **Hardware Access:** Setting the `mount_devices_into_container` key of `brick_config.yaml` to `true` instructs the orchestrator to automatically mount any physical hardware devices matching the classes in `required_devices` (e.g., `/dev/video0`) into the brick's Docker container.
```

Specify that these are referring to `brick_config.yaml`.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+- **Virtual Compose Network:** When defining custom services inside a standard [Docker Compose file](https://docs.docker.com/reference/compose-file/) named `brick_compose.yaml`, the orchestrator runs those containers inside an isolated virtual network. The custom brick's Python code running inside the main application container communicates with these services via network APIs (HTTP, WebSockets, or TCP/IP) using the compose service name as the hostname.
+- **Pre-compiled Images:** Arduino App Lab doesn't support compiling custom Docker images directly on the board from a `Dockerfile` via the `build` directive in `brick_compose.yaml`. You must build custom images beforehand, publish them to a public registry (such as GitHub Container Registry or Docker Hub), or load them onto the board manually using `docker load`.
```
per1234:
```suggestion
- **Custom Images:** Arduino App Lab doesn't support building custom Docker images directly on the board from a `Dockerfile` via the `build` directive in `brick_compose.yaml`. You must build custom images beforehand, publish them to a public registry (such as GitHub Container Registry or Docker Hub), or load them onto the board manually using `docker load`.
```

Correct terminology.

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Official Repositories
+
+| Repository | Purpose |
+| :---- | :---- |
+| `arduino/app-bricks-py` | Source code for preinstalled official bricks. |
+| `arduino/app-bricks-examples` | Reference applications demonstrating brick integration. These are accessible in Arduino App Lab. |
+| `arduino/arduino-app-lab` | The Arduino App Lab application. |
+| `arduino/arduino-app-cli` | Command-line tool for managing and validating apps. |
```
per1234:
```suggestion
| [`arduino/app-bricks-py`](https://github.com/arduino/app-bricks-py) | Source code for preinstalled official bricks. |
| [`arduino/app-bricks-examples`](https://github.com/arduino/app-bricks-examples) | Reference applications demonstrating brick integration. These are accessible in Arduino App Lab. |
| [`arduino/arduino-app-lab`](https://github.com/arduino/arduino-app-lab) | The Arduino App Lab application. |
| [`arduino/arduino-app-cli`](https://github.com/arduino/arduino-app-cli) | Command-line tool for managing and validating apps. |
```

per1234 (changes requested):

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
per1234: >do you think we should include some real examples?

Great idea! I think that would be very helpful to the reader.

> For example, the one you shared [here](https://forum.arduino.cc/t/uno-q-exploring-custom-bricks-and-app-lab-service-containers/1442898/2)

That Compose file was intended only for use directly with Docker for troubleshooting. It doesn't provide an interface that can be used by the Python script of an App.

I suggest providing two demonstrations, one for each fundamental concept:

- Use of a container
- Giving the container access to system resources

## Use of a container

```yaml
services:
  hello_server:
    # Create a container from a Docker image that provides the BSD Netcat networking utility.
    image: toolbelt/netcat
    # Use Netcat to create a web server that responds with a simple message.
    # See: https://manpages.debian.org/unstable/netcat-openbsd/nc.1.en.html
    entrypoint: |
      sh -c ' \
        while true; do
          echo -e "HTTP/1.1 200 OK\r\n\r\nHello, world!" \
          | \
            nc -l -p 5000 -q 0 -v
        done \
      '
    ports:
      # Expose the port for communication with the container's web server.
      # See: https://docs.docker.com/reference/compose-file/services/#ports
      - 5000
```

This Brick provides an interface that can be used from the Python code running on the App's primary container:

```py
import requests

from arduino.app_utils import App

response = requests.get("http://hello_server:5000")
print("Web server response:", response.text)

App.run()
```

Here is a complete demonstration App:

[simple-web-server-brick.zip](https://github.com/user-attachments/files/29425430/simple-web-server-brick.zip)

## Giving the container access to system resources

```yaml
# See: https://docs.docker.com/reference/compose-file/services/
services:
  sound_devices:
    # Create a container from a Docker image that provides the BSD Netcat networking utility.
    image: toolbelt/netcat
    # Make the host system's sound devices available for use inside the container.
    devices:
      - /dev/snd
    # Use Netcat to create a web server that responds with a list of the sound devices present in the container.
    # See: https://manpages.debian.org/unstable/netcat-openbsd/nc.1.en.html
    entrypoint: |
      sh -c ' \
        while true; do
          sound_devices="$$(ls -1 /dev/snd)"

          echo -e "HTTP/1.1 200 OK\r\n\r\n$$sound_devices" \
          | \
            nc -l -p 5000 -q 0 -v
        done \
      '
    ports:
      # Expose the port for communication with the container's web server.
      # See: https://docs.docker.com/reference/compose-file/services/#ports
      - 5000
```

Minimal Python code to interact with the Brick from the App's primary container:

```py
import requests

from arduino.app_utils import App

# Make an HTTP request to the custom Brick's web server.
# See: https://requests.readthedocs.io/en/latest/user/quickstart/#make-a-request
response = requests.get("http://sound_devices:5000")
print("Sound devices:")
print(response.text)

App.run()
```

Demonstration App:

[simple-device-access-brick.zip](https://github.com/user-attachments/files/29425418/simple-device-access-brick.zip)

content/software/app-lab/5.bricks/2.use-bricks/use-bricks.md
swikstrom: Great points, thanks!

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: Thanks for confirming!

content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md
swikstrom: @per1234 arduino-app-cli [validates the emoji here](https://github.com/swikstrom/arduino-app-cli/blob/babddae79207618e856adb89d957b4cb4623ecfc/internal/orchestrator/app/parser.go#L130-L143), using the [go.emoji](https://github.com/Andrew-M-C/go.emoji) library. That library appears to reliably parse emoji modifiers (like [skin tone](https://github.com/Andrew-M-C/go.emoji/blob/master/internal/official/emoji-sequences.txt#L1022-L1026)) correctly. In conclusion, I think there are no known restrictions 👍

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: Thanks Per! These examples are great. My learning preference is definitely to modify and experiment rather than read documentation and implement from scratch. I'm sure many users feel the same way 😄

I added a new ["Custom Brick Examples" section](https://github.com/arduino/docs-content/pull/2944/changes#diff-5e8ecd28164c65936ad71280fc435b17855f23e010012d01e40fd43aa8c3b74eR155-R246).

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,276 @@
...
+            nc -l -p 5000 -q 0 -v
+        done \
+      '
+    ports:
+      # Expose the port for communication with the container's web server.
+      # See: https://docs.docker.com/reference/compose-file/services/#ports
+      - 5000
+```
+
+This Brick provides an interface that can be used from the Python code running on the App's primary container:
```
swikstrom: @per1234 [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) seems to use "primary container" and "main container" interchangeably. Given the `main.py` filename, I think that "main container" would be a more consistent terminology.

```suggestion
This Brick provides an interface that can be used from the Python code running on the App's main container:
```

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,276 @@
...
+            nc -l -p 5000 -q 0 -v
+        done \
+      '
+    ports:
+      # Expose the port for communication with the container's web server.
+      # See: https://docs.docker.com/reference/compose-file/services/#ports
+      - 5000
+```
+
+Minimal Python code to interact with the Brick from the App's primary container:
```
swikstrom:
```suggestion
Minimal Python code to interact with the Brick from the App's main container:
```

@per1234 Same as for https://github.com/arduino/docs-content/pull/2944/changes#r3493608128.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
per1234: "Main container" is fine with me.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
```diff
@@ -0,0 +1,276 @@
...
+
+# Start the App
+App.run()
+```
+
+For more details on how to import and initialize Bricks, see [Use Bricks in Your App](../use-bricks/).
+
+## Custom Brick Examples
+
+These examples illustrate key custom container patterns. Each example includes a complete, downloadable App archive that you can [import into Arduino App Lab](../../apps/manage-apps/#import-an-app).
```
per1234:
```suggestion
These examples illustrate key custom container concepts. Each example includes a complete, downloadable App archive that you can [import into Arduino App Lab](../../apps/manage-apps/#import-an-app).
```

"Pattern" indicates that these could be considered approaches that would be commonly built upon when developing more complex custom Bricks. I'm not sure these could be considered as such. My goal was to produce minimal "Hello, world!" type of demonstrations. It is possible to adopt this approach to some applications, but it is going to be awkward to implement anything complex via the "entrypoint" command alone.

Ideally we would also provide a demonstration of an approach to a custom Brick that is better suited to building on. I would suggest the use of a [Flask](https://flask.palletsprojects.com/)-based web server, something like what is done in this minimal App:

https://github.com/user-attachments/files/27430584/Custom.Brick.With.Container.zip

However, things become significantly more complicated once you move into creating a custom image source, as you explain here:

https://github.com/arduino/docs-content/blob/f66a76c13b0e0ba11788332658638586f4a84eae/content/software/app-lab/5.bricks/4.bricks-reference/bricks-reference.md?plain=1#L136

I think it would be necessary to provide a full tutorial for managing containers built from a custom image source (perhaps something like what is described [here](https://github.com/mjs513/Arduino-Uno-Q-SPI3-Bridge-App/blob/main/README.md#automatic-rebuilds-with-docker-compose-watch) per [this discussion](https://forum.arduino.cc/t/getting-a-neato-xv-11-lidar-working-on-the-q/1445568/31)) along with such a demonstration App.

per1234 (changes requested):

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: @per1234 Thanks again! To avoid further expanding the scope of this PR, I'll commit the suggested change only, so we can go ahead with the merge. I've created [PXCT-2036](https://arduino.atlassian.net/browse/PXCT-2036) to track the addition of a dedicated tutorial. You're welcome to comment or reach out directly if you find any other useful examples or resources.

content/software/app-lab/5.bricks/3.custom-bricks/custom-bricks.md
swikstrom: Related: https://github.com/arduino/docs-content/pull/2944#discussion_r3511794419

per1234 (approved):
Thanks Sebastian!

PR #2955, opened 2026-05-13
[PXCT-1948] UNO Q: Shutdown Instructions Update

TaddyHC:
## What This PR Changes
- Updates the shutdown instructions found within the UNO Q Debian documentation

## Contribution Guidelines

### DIFF
--- a/content/hardware/02.uno/boards/uno-q/tutorials/07.debian-guide/content.md
+++ b/content/hardware/02.uno/boards/uno-q/tutorials/07.debian-guide/content.md
@@ -1180,7 +1180,7 @@ To safely power down your UNO Q for extended storage or when carrying the board,
 sudo halt
 ```
 
-The `sudo halt` command stops all Linux processes and brings the system to a safe halted state. The board remains powered on, but Linux stops running, the screen goes black, and the system becomes unresponsive to network connections.
+The `sudo halt` command stops all Linux processes and brings the system to a safe halted state. The board remains powered on, but Linux stops running and the system becomes unresponsive to network connections.
 
 It is different from a complete power-off but represents a safe state in which all files are properly closed and the filesystem is protected.
 
@@ -1196,7 +1196,13 @@ sudo halt
 
 ![Shutting Down Your UNO Q Safely (1)](assets/debian_shutdown_halt.gif)
 
-After running this command, watch for the screen to go black, indicating the system is in the halt process. Wait approximately 10 seconds after the screen goes black to make sure the halt is complete. The board will remain powered on but in a halted state. Power can then be safely disconnected; the filesystem has closed safely and all processes have stopped.
+After running this command, the shutdown process begins. The visual indicators and timing depend on how you are connected to the board.
+
+If using SBC mode with a display connected to the board, watch for the screen connected to the board to go black, indicating the system is in the halt process. Wait approximately 10 seconds after the screen goes black to make sure the halt is complete. The board will remain powered on but in a halted state. Power can then be safely disconnected, as the filesystem has closed safely and all processes have stopped.
+
+If using PC-hosted mode (USB-C/ADB or SSH connection, no display on board), after running the `sudo halt` command, wait approximately 10-15 seconds to make sure the halt is complete. The board will remain powered on but in a halted state. Power can then be safely disconnected, as the filesystem has closed safely and all processes have stopped.
+
+If using Arduino App Lab, it may return to the `Welcome to Arduino App Lab` board discovery view after the shutdown completes.
 
 The `sudo halt` command is ideal for long-term storage, carrying the board, or any situation where you want the board to stay off until you manually power it back on.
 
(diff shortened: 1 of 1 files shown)


content/hardware/02.uno/boards/uno-q/tutorials/07.debian-guide/content.md
```diff
@@ -1192,10 +1196,30 @@ sudo halt
 
 ![Shutting Down Your UNO Q Safely (1)](assets/debian_shutdown_halt.gif)
 
-The green power LED on the board will turn off when the system has halted completely. The board will remain powered off and will not restart automatically.
+The green power LED on the board will turn off when the system has halted completely. The board will remain powered on but in a halted state. Power can be disconnected at any time after the halt completes, the filesystem is already safely closed and all processes have stopped.
```
per1234: > The green power LED on the board will turn off

Have you personally verified this statement to be true? It doesn't do that when I try it.

TaddyHC: Before making the updates for shutdown commands, I did verify the commands and behavior again. The LED matrix stays on as they do but seems the green LED behavior needs to be fixed. I may have confused it with intermittency from past. I'll make the correction and have it posted in a while

TaddyHC: I have pushed an update to update the corresponding information, thanks

per1234: The new "screen goes black" indicator is only applicable to the case where the user has a display connected to the board. The documentation does not communicate this use case-specificity of the information. The reader who is using the board in PC-hosted mode might assume it means something like the terminal or App Lab window going black and therefore be confused.

I suggest the following refinements:

- Adjust the wording of the statements to make it clear they are referring to the case where a display is connected to the board.
- Add an alternative indicator for the case where no display is connected.

As for the latter, I have found myself utilizing the blue [QRB 2 LED](https://docs.arduino.cc/tutorials/uno-q/user-manual/#mpu-controlled-leds) on the board as an indicator of the operating system state. However, this LED is apparently intended to be an indicator of Bluetooth activity, so I don't know if it is actually reliable for this purpose.

Alternative indicators could be:

- The shell session disconnects in the terminal. This applies for connection via ADB and SSH both.
- Arduino App Lab goes back to the "Welcome to Arduino App Lab" board discovery view. I'm not sure how reliable this is though.

TaddyHC: Thanks for the heads up, that's right, the "screen goes black" indicator only applies when a display is physically connected to the board in SBC mode. The documentation is now updated to distinguish between SBC mode with a display connected to the board, where you use the screen going black as the indicator, and PC-hosted mode using USB-C, ADB or SSH without a display, where you wait 10-15 seconds after the command.

Regarding the QRB2 LED, I tested it and found it's not reliable for part of shutdown/halt indication. It turns off when the command is entered and turns blue again if auto-restart occurs which could be used in this specific instance. However, if an App is running, the QRB2 LED transitions to and stays off, and it would also depend on the application. When I enter the halt command with an App running, the LED provides no indication, including the quick green blink that normally is observable when the board runs commands without an App. For PC-hosted mode, we can use the time-based approach of waiting 10-15 seconds. The documentation now describes which indicator applies to each connection mode.

content/hardware/02.uno/boards/uno-q/tutorials/07.debian-guide/content.md
```diff
@@ -1180,9 +1180,13 @@ To safely power down your UNO Q for extended storage or when carrying the board,
...
 ```
 
+The `sudo halt` command stops all Linux processes and brings the system to a safe halted state. The board remains powered on, but Linux stops running and the system becomes unresponsive to network connections.
+
+It is different from a complete power-off but represents a safe state in which all files are properly closed and the filesystem is protected.
+
 #### Recommended Shutdown Method
 
-The `halt` command is the recommended approach for safely shutting down your UNO Q. It stops all system processes, brings Linux to a safe state, and keeps the board powered off, preventing an automatic restart.
+The `halt` command is the recommended approach for safely shutting down your UNO Q. It stops all system processes, brings Linux to a safe state, and in most cases with current firmware, keeps the board powered off without triggering an automatic restart.
```
per1234:
```suggestion
The `halt` command is the recommended approach for safely shutting down your UNO Q. It stops all system processes, brings Linux to a safe state, and in most cases with current firmware, keeps the board in that state without triggering an automatic restart.
```

From the information provided later in the document, it is my understanding that `halt` only puts the operating system into a halted state, but leaves the hardware in a powered state.

I'm not able to verify the behavior either way because my boards always restart.

TaddyHC: Yes, that is correct. I have updated the relevant sections aligning with your suggestion. I see it's better given the varying behavior across different configurations. I tested again with my board, even after waiting 5+ minutes after the command, the board remains powered on but inaccessible unless I reconnect the power.

per1234 (commented):
Thanks @TaddyHC!

canchebagur (approved):
LGTM!

PR #2959, opened 2026-05-16
Fix formatting in Modulino Thermo tutorial content

n3orma:
@jhansson-ard


g-castronovo (approved):
Thank you - I double checked the urls :)) works now

PR #3012, opened 2026-08-04
[PXCT-2107] UNO Q: Datasheet JANALOG Detail Update

TaddyHC:
## What This PR Changes
- Updates UNO Q datasheet regarding the 5V pin on JANALOG header

## Contribution Guidelines


g-castronovo:
Mhh but I need to check this one too right?

mcmchris (approved):
LGTM! Btw, I would like to know which are the implications of powering through the 5V Pin, because I don't see any technical limitation.

TaddyHC:
@g-castronovo yes, now that we have the review, we can move forward with the corresponding translations needed for the updated details :)

@mcmchris alright, let's hold off for now just in case, so we have all the information up to date and we can make the addition if needed :)

jhansson-ard:
We should wait with merging this PR until @g-castronovo's PR is merged for the translations

jhansson-ard:
@g-castronovo @TaddyHC does the new UNO Q datasheets include this fix?

TaddyHC:
Hi @jhansson-ard , the new datasheet doesn't include this yet. Given the reference thread, though, here is where it stands: 5V_SYS is the correct net on JANALOG, JMISC and JSPI (the diode-OR output), not 5V_USB_VBUS as currently written. We also got confirmation from the thread that powering the board from 5V_SYS on any of those three headers is fine, within latest schematics as well that would be updated if I'm not wrong. I'm putting together a PR with these updates now

TaddyHC:
Here is the latest PR updated on top of the latest doc updates https://github.com/arduino/docs-content/pull/3025

I'll have it posted once ready for review, running through the final checks on the JANALOG/JMISC/JSPI update

Cc. @jhansson-ard @g-castronovo @mcmchris

PR #3004, opened 2026-07-08
Clean up old branches monthly

Matteo-it:
Adds a GitHub Action `stale-branch-cleanup.yml` that automatically removes stale remote branches. The repo has accumulated many old branches (stale, closed, already merged), and this workflow cleans them up safely.

**What it does**

- Runs automatically once a month (03:00 UTC, day 1) via schedule cron.
- Deletes remote branches whose last commit is older than 365 days.
- Can also be triggered manually via workflow_dispatch.

**Safety**

- Always skips the default branch and common protected names (main, master, develop, staging, production, gh-pages).
- Dry-run mode: manual runs default to a simulation that only lists candidate branches without deleting. Scheduled runs perform the actual deletion.
- Configurable days_threshold on manual runs.
- Uses only GITHUB_TOKEN with contents: write, no extra secrets.Adds a GitHub Action ([stale-branch-cleanup.yml](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) that automatically removes stale remote branches. The repo has accumulated many old branches (stale, closed, already merged), and this workflow cleans them up safely.

PR #3022, opened 2026-08-25
[2.2.9] Improve build time

Matteo-it:
Speed up deploy builds and stop stalled checkouts from hanging


.github/workflows/deploy-staging.yml
```diff
@@ -117,36 +196,68 @@ jobs:
...
+          xargs -0 -r -P "$(nproc)" -n 1 jpegoptim --strip-all < /tmp/big-jpg.lst || echo "Some JPGs could not be optimized, continuing"
           echo "Compressing large PNGs..."
-          find content -type f -name "*.png" -size +2M -print -exec optipng -o2 {} \;
+          xargs -0 -r -P "$(nproc)" -n 1 optipng -o2 < /tmp/big-png.lst || echo "Some PNGs could not be optimized, continuing"
+          # The list is captured before compressing: some files drop under 2M and find would no longer report them
+          cat /tmp/big-jpg.lst /tmp/big-png.lst | tar --null -cf /tmp/compressed-images.tar -T -
 
       - name: Gatsby main cache
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v5
```
damianoc90: This restores the Gatsby cache but nothing in this workflow saves it any more: `deploy-staging.yml` has two `cache/restore` steps and zero `cache/save`. Same in `preview.yml`. Only `deploy-production.yml` got the full restore+save pair.

`actions/cache@v4` saved automatically in the post-job, so before this change staging wrote its own `-cache-gatsby-ghaction` / `-public-gatsby-ghaction` on every run. Now that key is never written by anyone, so every staging deploy falls back through `restore-keys` to the `…-node22-main` prefix — i.e. to whatever the last **production** build wrote, built with `APP_ENV=prod`. Staging has effectively lost its own warm cache, which cuts against what the PR is for.

If that is deliberate (only production pays the ~4 GB of quota), it needs saying here: the comment explaining who writes the cache lives only in deploy-production.yml, where it reads as a statement about branches rather than about workflows. Otherwise this wants the two `cache/save` steps mirrored over from production.

.github/workflows/preview.yml
```diff
@@ -120,34 +190,60 @@ jobs:
...
+          xargs -0 -r -P "$(nproc)" -n 1 jpegoptim --strip-all < /tmp/big-jpg.lst || echo "Some JPGs could not be optimized, continuing"
           echo "Compressing large PNGs..."
-          find content -type f -name "*.png" -size +2M -print -exec optipng -o2 {} \;
+          xargs -0 -r -P "$(nproc)" -n 1 optipng -o2 < /tmp/big-png.lst || echo "Some PNGs could not be optimized, continuing"
+          # The list is captured before compressing: some files drop under 2M and find would no longer report them
+          cat /tmp/big-jpg.lst /tmp/big-png.lst | tar --null -cf /tmp/compressed-images.tar -T -
 
       - name: Gatsby main cache
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v5
```
damianoc90: Same as in deploy-staging.yml: `restore` with no `save` anywhere in the file, where `actions/cache@v4` used to persist it in the post-job.

The cost lands squarely on the iterative case this workflow exists for. The primary key is `…-node22-${{ github.ref_name }}`, which on a `pull_request` event is `<number>/merge`: it used to be written at the end of every preview build, so the second and later pushes on the same PR restored the previous build's `.cache` and `public/` exactly. Now nothing writes it and each push restarts from `main`'s cache — and the `preview` label flow is precisely where people push repeatedly and wait for the deploy.

`public/` is the expensive half (the comment in production puts it at ~38 minutes of sharp), so this is worth measuring before deciding it is fine.

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "homepage": "https://github.com/arduino/docs-content#readme",
   "dependencies": {
-    "@arduino/docs-arduino-cc": "^2.2.81",
+    "@arduino/docs-arduino-cc": "2.2.9",
```
damianoc90, line 22: This reads as a downgrade of about 72 patch releases, not a bump: in semver `2.2.9 < 2.2.81`. The lockfile confirms both are real published releases — the resolved URL and the integrity hash both change — so npm is not going to complain either way.

The PR title says "[2.2.9]" and one of the commits is "fix: version", which is what makes me ask: was the intent to move **up** to a `2.2.9x` release, or is this package's numbering genuinely non-semver (a `2.2` series with a release counter, where 9 came after 81)? Dropping the `^` in the same edit fits the second reading, since a caret range cannot express that ordering.

It is a private GitHub Packages package, so I cannot check the publish dates from outside — you can. If the ordering really is inverted, a short comment next to the pin would save the next reader the same detour; if it is not, this silently ships an older theme to production.

.github/workflows/deploy-production.yml
```diff
@@ -117,36 +196,75 @@ jobs:
...
 
       - name: Build
         run: npm run build
+        env:
+          # Gatsby defaults to physical cores, which is half the vCPUs of a runner
+          GATSBY_CPU_COUNT: logical_cores
+
+      # A warm public/ is worth 38 minutes of sharp, and costs ~4 GB of quota: main writes it, every other branch only reads
+      - name: Save Gatsby main cache
+        uses: actions/cache/save@v5
```
damianoc90, line 266: The key hardcodes `main` and there is no branch guard, while the comment above says "main writes it, every other branch only reads". This workflow has `workflow_dispatch`, so it can run from any ref, and a dispatch from branch X would write a cache literally named `…-node22-main-<run_id>`.

GitHub's cache scoping keeps that from reaching `main` — a cache created on a branch is visible only to that branch — so this is not a poisoning bug. But X then keeps restoring its own entry through the `-node22-main` prefix instead of main's, and the ~4 GB the comment mentions gets spent per dispatching branch rather than once.

`if: github.ref_name == 'main'` on both save steps makes the code say what the comment already claims.

.github/workflows/stale-branch-cleanup.yml
```diff
@@ -0,0 +1,270 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Process stale branches
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+          # For scheduled runs the inputs are empty, so fall back to defaults.
+          DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
```
damianoc90: The declared dispatch default is `true`, but the fallback here is `'false'`, so the **scheduled** run deletes branches and closes pull requests for real. The comment says "fall back to defaults", which is what tripped me up: the default it falls back to is the opposite of the input's.

That may well be what you want long-term, but it means the first cron after merge acts with no rehearsal. A dispatch with `dry_run=true` first, and reading the report, is cheap insurance — the script is careful (protected-branch regex, base branches of open PRs, cross-check against the activity log, exempt label, reminder cycle before closing), so I expect it to behave; I would just rather see it say so once before it starts deleting.

Two smaller things while here: the cron is `0 3 * * 1`, i.e. **weekly**, while the commit that brought this in says "monthly" — worth settling which is intended. And this workflow is unrelated to build time and unmentioned in the PR description; a 270-line job with `contents: write` that deletes refs deserves a reviewer looking at it as its own change, not as a rider on a CI-speed PR.

.github/workflows/deploy-production.yml
```diff
@@ -117,36 +196,75 @@ jobs:
       #       echo "No oversized files found. New check passed."
       #     fi
 
+      # Key on the paths and sizes of the files the step compresses, so it changes as soon as any of them does
+      - name: Compressed images key
+        id: images-key
```
damianoc90, line 201: The key hashes paths and **sizes**, so two different images of identical byte size are indistinguishable to it. Replace one big JPEG with another that happens to weigh exactly the same and the key does not move: the cache hits, `tar xf` restores the *previous* image's compressed bytes over the new file, and the build ships the wrong picture with nothing in the log to suggest it.

Unlikely, but silent and about as hard to debug as a caching bug gets — and the fix is the same length: hash the contents (`hashFiles` over the same globs, or piping the file list through `git hash-object --stdin-paths`) instead of `%s`.

Same step in deploy-staging.yml and preview.yml. Everything else about this cache checks out, for what it's worth: the file list is captured before compressing, the tar paths are workspace-relative, and a hit short-circuits before the apt install.

content/hardware/02.uno/boards/uno-q/datasheet/it-datasheet.md
```diff
@@ -997,9 +997,8 @@ Le guide d’utilisation des dispositifs pour réseaux locaux doit inclure des i
 
 (i) les dispositifs fonctionnant dans la bande 5 150-5 250 MHz sont réservés uniquement pour une utilisation à l’intérieur afin de réduire les risques de brouillage préjudiciable aux systèmes de satellites mobiles utilisant les mêmes canaux ;
 
-<div style="background-color: rgba(0, 170, 228, 0.2); border-left: 6px solid rgba(0, 120, 180, 1); margin: 20px 0; padding: 15px;">
-  <strong>Note:</strong> For 5GHz and/or when co-located with 5 GHz transmitters, the following statements should be provided in the user information. 
-</div>
+> [!NOTE]
```
damianoc90, line 1000: This change and the `blockquote.warning` / `blockquote.caution` rules in `common-style.css` both depend on the bumped `datasheet-renderer` commit turning `[!NOTE]` into a styled blockquote — and nothing in this PR has exercised that: `Render Datasheets` reports *skipping* on every run here, alongside `preview-build`. If the renderer does not special-case the marker, the literal text `[!NOTE]` ends up in the PDF.

One dispatch of `render-datasheets.yaml` and a look at the artifact would settle it before this reaches a published datasheet.

Minor, in the same block: the `**Note:**` inside the callout is now redundant with the `[!NOTE]` marker, so it will likely read "Note" twice.

damianoc90 (commented):
Reviewed the whole diff. The build work is well argued — the retry ladder wipes the workspace between attempts, every step is bounded and the jobs now have their own timeouts, and the `tj-actions/changed-files` pin is honest: I checked that `9426d409…` really is the commit behind `v47.0.6`.

Three things I would fix before merge (first two inline below, plus the cache-save guard), and a few worth a look.

Not worth their own threads:

- **`npm install` -> `npm ci`.** For a lockfile-driven deploy `ci` is the right verb and is usually faster with a warm npm cache; `install` can rewrite the lockfile in CI.
- **Action majors.** They all exist, but several are one or two behind: checkout v5 (latest v7), setup-node v5 (v7), cache v5 (v6). `upload-artifact@v7`/`download-artifact@v8` are current. If the lag is deliberate, fine — but this is the PR that touches all of them.
- **`xargs … || echo "continuing"`** swallows a systemic failure too, not just the handful of .png files that are really JPEGs. If `optipng` fails to install, the build proceeds with uncompressed images and says so only in a log line.
- The third checkout attempt has no `timeout-minutes`. Fine as it stands, since the new job timeout bounds it — worth a word in the comment so it reads as a choice.

.github/workflows/deploy-staging.yml
Matteo-it: You're right on both counts: this file lost its own cache, and the policy comment lived only in `deploy-production.yml`. Both fixed — staging goes back to `actions/cache@v5` (restore plus the post-job save) rather than mirroring production's explicit save pair.

I kept it as the plain action instead of a restore/save pair because the quota measurement points at the previews, not at staging. When main's `public/` entry got evicted, the two entries holding the 10 GB were `Linux-public-gatsby-3033/merge` and `Linux-public-gatsby-3036/merge`, 3.9 GB each: staging writes one entry per branch under test, previews write one per open PR. So staging keeps its warm cache, and the file now says who writes what.

.github/workflows/preview.yml
Matteo-it: Agreed on the cost, and it is measurable: a build's `public/` entry is ~3.9 GB while `.cache` is ~0.1 GB.

So preview now saves `.cache` with `actions/cache@v5` — each PR keeps it across pushes, which is the iterative case this workflow exists for — and still only restores `public/`. One 3.9 GB entry per open PR is what filled the 10 GB quota and evicted `…-node22-main`, and a miss on that prefix is the ~29-38 minutes of `gatsby-plugin-sharp` you quote (measured at 1747 s and 2285 s in two runs). Keeping `public/` per PR as well needs a bigger quota first, not a smaller `public/`.

package.json
Matteo-it: Your second reading is the correct one: the numbering is a release counter rather than semver, and the ordering really is inverted — `2.2.9` was published after `2.2.81`. From the inside the publish dates confirm it.

That is also why the `^` had to go in the same edit: `semver.maxSatisfying(['2.2.7','2.2.8','2.2.81','2.2.9'], '^2.2.9')` returns `2.2.81`, so a caret range silently ships the *older* theme. It already happened: the auto-generated bump PRs on `docs-content-reference` and `docs-content-library` had `^2.2.9` in `package.json` and `2.2.81` resolved in their lockfiles, until they were pinned exactly.

A comment next to the pin is not possible in JSON, so the explanation goes in the PR description instead. The way out is the next release, and it has to clear 2.2.81: `2.2.10` does not (10 < 81), `2.3.0` does — after that all three repos can go back to carets.

.github/workflows/deploy-production.yml
Matteo-it: Agreed, fixed: `if: github.ref_name == 'main'` on both save steps, so the code now says what the comment claims.

One consequence worth writing down, since it changes the plan for verifying this PR: a pre-merge dispatch of this workflow from the feature branch will no longer populate `…-node22-main`. That entry now appears with the first production run after merge, and the last staging run before it still pays the sharp phase. Correct, but it is a step we were counting on.

.github/workflows/stale-branch-cleanup.yml
Matteo-it: Agreed on both, and the fallback is worse than it looks: `github.event.inputs.dry_run || 'false'` means the cron acts for real on its very first firing, while a manual dispatch defaults to a rehearsal.

I ran the selection locally against the current repo: 147 remote branches, 109 with no commits for over a year, of which **71 would be deleted outright** and **38 would start the reminder cycle** on their open PRs, with 1 protected as the base of an open PR. So a dispatch with `dry_run=true` and a read of the summary table is cheap insurance before the first cron.

You are also right that it does not belong in this PR: it arrived here as part of a merge from `main` (#3004), not as part of the build work, and 270 lines with `contents: write` that delete refs deserve their own review.

The cron question is still open: the file says `0 3 * * 1`, i.e. weekly, which is what makes the three reminders at seven days close a stale PR in about three weeks rather than three months.

.github/workflows/deploy-production.yml
Matteo-it: Good catch, and fixed. I went with git rather than `hashFiles` so the hashing stays proportional to the 52 files the step actually touches instead of every image under `content/`:

```bash
find content -type f \( -name '*.jpg' -o -name '*.jpeg' -o -name '*.png' \) -size +2M -print0 > /tmp/big-images.lst
xargs -0 -r git ls-files -s -- < /tmp/big-images.lst > /tmp/big-images.idx
test "$(tr -dc '\0' < /tmp/big-images.lst | wc -c)" -eq "$(wc -l < /tmp/big-images.idx)"
echo "hash=$(sort /tmp/big-images.idx | sha256sum | cut -c1-40)" >> "$GITHUB_OUTPUT"
```

`git ls-files -s` prints mode, blob hash and path, so the key moves on any content change and on a rename, and costs nothing because git already has the hashes.

The `test` closes the hole your objection opens from the other side: `ls-files` reads the index, so an untracked image would be compressed but absent from the key. Now the counts have to match or the job fails. Verified locally: 52 files, 52 index lines.

content/hardware/02.uno/boards/uno-q/datasheet/it-datasheet.md
Matteo-it: Two separate points here, and I disagree with the second one.

**On the marker never having been exercised.** Right about this PR's checks — `Render Datasheets` is gated on the `preview` label and reports skipping. But the path has run: staging run `33487314779`, on commit `c621fcf34`, already carried the bumped pin (`datasheet-renderer.git#97e240e0`), and `generate-datasheets` ran there for **5.1 minutes** instead of hitting its cache, because editing `it-datasheet.md` changed the `hashFiles('**/*datasheet.md')` key. Every PDF in that run was re-rendered by the new renderer.

**On "Note" appearing twice.** That is GitHub's behaviour, not this renderer's. GitHub turns `[!NOTE]` into a titled callout; ours strips the marker and adds a class, injecting no label. Rendering that exact file through the installed module gives:

```html
<blockquote class="note">
<p><strong>Note:</strong> For 5GHz and/or when co-located with 5 GHz transmitters, the following statements should be provided in the user information.</p>
</blockquote>
```

One "Note", the one in the text. Removing it would leave the box unlabelled, which is a regression against the `<div>` it replaces.

Your closing ask stands, though, and it is the one thing still missing: **nobody has looked at the PDF**. Local rendering does not work here (puppeteer's Chromium crashes on macOS), so my verification is HTML through the real `marked` and the real stylesheet. A dispatch of `render-datasheets.yaml` and a look at the artifact is the check I want before this reaches a published datasheet — in particular to see how the box behaves across a page break, which the empty `<div>` never had to.

Matteo-it:
Thanks — the four items outside the threads, in order.

**`npm install` -> `npm ci`.** Deliberate, and I would keep `install`. `ci` installs the lockfile exactly, which freezes `@arduino/docs-arduino-cc` at the pinned version: today a theme release reaches the sites because `install` re-resolves it, and with `ci` every release would need a lockfile bump in each consumer repo first. `npm ci` is the better verb for a lockfile-driven deploy in general — it just conflicts with how this project ships the theme. The speed argument is also weaker than it looks here: with the warm npm cache and `--prefer-offline --no-audit --fund=false` the step went from 2.5 min to 0.6-0.8 min, which is the win we were after.

**Action majors one or two behind.** Deliberate for this PR. We learned the cost the hard way: bumping Node 18 -> 22 added `node22` to the cache keys, which correctly invalidated every existing entry and cost one run 29 minutes of `gatsby-plugin-sharp`. Stacking three more major jumps into the same PR while we are measuring build time mixes the variables. Happy to do them as a follow-up, where a regression is attributable.

**`xargs … || echo` swallowing a systemic failure.** You are right, fixed. The install is now followed by `command -v jpegoptim > /dev/null && command -v optipng > /dev/null`, so a missing compressor stops the job while the per-file tolerance stays for the handful of `.png` files that are really JPEGs (three of them, all in `content/hardware/09.kits/maker/iot-bundle`, if anyone wants to fix them at the source).

**Third checkout attempt without `timeout-minutes`.** Reworded so it reads as a choice: "one more wipe and a last try, bounded by the job timeout".

package.json
Matteo-it: Closing the loop from the other side: the decision is to keep `2.2.9` rather than renumber, so the exact pin in this PR stays as the deliberate workaround rather than a temporary one, and the explanation goes in the PR description as agreed.

The one thing to carry forward is that `docs-auto-bump.yml` writes `"^${VERSION}"`: with this numbering a caret resolves back to `2.2.81`, which is how the reference and library bump PRs ended up pinning the old theme in their lockfiles. Whoever runs the next bump needs an exact pin, or that workflow needs changing first.

.github/workflows/stale-branch-cleanup.yml
damianoc90: Re-checked at 0df8e21dd: the file is unchanged. `cron: '0 3 * * 1'` and `DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}` are both still there, so the mitigation we agreed on lives in this thread and not in the workflow.

That matters because merging is what arms it. On the first Monday after merge the job runs unattended with `dry_run=false`, and by your own simulation that is 71 branches deleted outright and the reminder cycle opened on 38 open pull requests — no rehearsal, no summary read first.

The shape of it is backwards, and that is the part I would fix rather than work around: a human who dispatches deliberately gets a rehearsal by default, while the unattended path gets the destructive one. The conservative default belongs on the path nobody is watching.

Either of these is enough, and both are small:

1. Flip the fallback to `'true'` now, dispatch once, read the summary, then flip it to `'false'` in a one-line follow-up. The first real run then happens with someone looking at it.
2. Leave the code as it is and do the dry-run dispatch **before** this merges rather than after, so the 71 deletions are reviewed while the cron is still disarmed.

Since the file arrived here from the #3004 merge and is not part of the build work, option 1 also has the advantage of not making this PR the thing that arms it.

The cron question is still open too, and it is not cosmetic: at `0 3 * * 1` the three reminders at seven-day intervals close a stale pull request in about three weeks. If the intent was monthly — as the commit that brought this in says — the interval constants want revisiting alongside the schedule, or the reminder window ends up far shorter than "we will close this if nothing happens" implies to the person receiving it.

Matteo-it: You are right, and I was wrong on the fact my previous reply rested on.

I wrote that the file "arrived here as part of a merge from `main` (#3004)". It did not. #3004's base was `Matteo-it/improve-build-time`, not `main`:

```
$ gh api repos/arduino/docs-content/pulls/3004 --jq '.base.ref'
Matteo-it/improve-build-time

$ git show origin/main:.github/workflows/stale-branch-cleanup.yml
fatal: path '.github/workflows/stale-branch-cleanup.yml' does not exist in 'origin/main'
```

So the workflow does not exist on the default branch, is not registered in the repository at all, has never fired, and **this PR is what puts it on `main` and starts the cron**. My framing minimized exactly the risk you are pointing at, and the mitigation did belong in the file rather than in this thread.

**Your option 2 is not available**, which settles the choice between the two. `workflow_dispatch` requires the workflow to be present on the default branch, so the rehearsal cannot be run before the merge:

```
$ gh workflow run stale-branch-cleanup.yml --ref Matteo-it/improve-build-time -f dry_run=true
HTTP 404: workflow stale-branch-cleanup.yml not found on the default branch
```

**Fixed — but not with the literal flip to `'true'`, because that carries the same bug mirrored.** With `type: boolean`, unchecking the box sends boolean `false`, which is falsy, so `|| 'true'` would fire and turn a deliberate real run into a rehearsal. The knob would look like it worked and quietly do nothing, which is the failure mode you spotted, pointed the other way. Keying on the event instead:

```yaml
# A cron has nobody watching it, so it only reports unless a person dispatches and asks for the real run
DRY_RUN: ${{ github.event_name != 'workflow_dispatch' || github.event.inputs.dry_run }}
```

| event | expression | `DRY_RUN` |
|---|---|---|
| `schedule` | `true \|\| …` | `true` — reports only |
| dispatch, box checked | `false \|\| true` | `true` — reports only |
| dispatch, box unchecked | `false \|\| false` | `false` — real run |

The consequence, stated plainly rather than left implicit: **the cron is now inert.** It will report every Monday and delete nothing until someone changes that line on purpose. That is the state I want it in for this merge — arming an unattended job that deletes 71 refs should be a deliberate one-line commit after a summary has been read, not a side effect of a build-time PR.

The schedule question is still open and still not cosmetic: `0 3 * * 1` is weekly, while the commit that brought this in says "monthly". With `MAX_REMINDERS: 3` at `REMINDER_INTERVAL_DAYS: 7`, one reminder per run, a stale pull request closes about three weeks after the first notice. If monthly was the intent, the interval constants want revisiting together with the cron, or the window ends up much shorter than "we will close this if nothing happens" sounds to whoever receives it.

damianoc90 (approved):

swikstrom:
Looks great.  Because `fetch-depth: 1` blanks out `gitLogLatestDate`, the previews will have inaccurate "last updated" dates, but given your comments, this is an intentional trade-off to increase the speed even further. Makes sense to me 👍

@Matteo-it I see you just merged the PR, but I'll post this anyway just to say thanks!

